PR #3, opened 2017-11-09
Clarify the progress of DLEQ proof implementations

alxdavids:


alxdavids:
out-of-band approval from Nick, so merging.

Issue #2, opened 2017-11-09
DLEQ proof verification is not finished


alxdavids:
Currently, we do not recompute the PRNG state within the browser extension when verifying the batch proofs. We need to implement support for SHAKE-256 (which is not available currently in the SJCL library) in order to do this. 

- The server at the moment sends the value C that is computed as the output of PRNG(H(G,Y,M[i],Z[i])).
- The extension should recompute C but instead we use the value of C that is sent by the server.
- This is incorrect and should be fixed before the DLEQ proofs are deemed complete.

alxdavids:
This has been fixed now. We still regard the protocol/extension as in beta development and would appreciate more scrutiny on both! Closing this issue for now though.

PR #12, opened 2017-11-20
Fix DLEQ proof verification

alxdavids:
- Proof computation was incorrect (Issue #2 )
-- We now explicitly recompute PRNG (SHAKE256) seed
-- Do not check unnecessary proof values sent by server
- Include browserify'd version of keccak npm module for SHAKE256 (compatible with challenge-bypass-server)
- Updated readme to be consistent with updated verification
- Updated manifest version to v1.0.1
- Tested receiving signed tokens in Chrome & Firefox successfully

(also more reviewers welcome)

### DIFF
--- a/scripts/crypto.js
+++ b/scripts/crypto.js
@@ -19,6 +19,7 @@
 
 var p256 = sjcl.ecc.curves.c256;
 const BATCH_PROOF_PREFIX = "batch-proof=";
+const UNCOMPRESSED_POINT_PREFIX = "04";
 const MASK = ["0xff", "0x1", "0x3", "0x7", "0xf", "0x1f", "0x3f", "0x7f"];
 
 const DIGEST_INEQUALITY_ERR = "[privacy-pass]: Recomputed digest does not equal received digest";
@@ -444,6 +445,6 @@ function getBigNumFromHex(hex) {
 // PRNG encode point
 function encodePointForPRNG(point) {
     let hex = sjcl.codec.hex.fromBits(point.toBits());
-    let newHex = "04" + hex;
+    let newHex = UNCOMPRESSED_POINT_PREFIX + hex;
     return sjcl.codec.hex.toBits(newHex);
 }
--- a/scripts/crypto.js
+++ b/scripts/crypto.js
@@ -14,23 +14,18 @@
 /* exported sec1DecodePoint */
 /* exported signPoint */
 /* exported unblindPoint */
-/* exported verifyBatchProof */
+/* exported verifyProof */
 "use strict";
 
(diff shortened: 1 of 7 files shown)


scripts/crypto.js
```diff
@@ -479,38 +419,31 @@ function getMarshaledBatchProof(proof) {
...
+
+// Return a bignum from a hex string
+function getBigNumFromHex(hex) {
+    return sjcl.bn.fromBits(sjcl.codec.hex.toBits(hex));
+}
+
+// PRNG encode point
+function encodePointForPRNG(point) {
+    let hex = sjcl.codec.hex.fromBits(point.toBits());
+    let newHex = "04" + hex;
```
lziest: is there anything special about this prefix?

scripts/crypto.js
```diff
@@ -353,120 +339,74 @@ function verifyDleq(bp, chkM, chkZ) {
...
-        console.error(INCONSISTENT_DLEQ_PROOF_ERR);
-        return false;
-    }
-    return true;
+
+    return {M: cM, Z: cZ};
 }
 
-// Checks that the batch proof is well-defined
-function isBatchProofCompleteAndSane(bp, chkM, chkZ) {
```
lziest: Nice. This gist of this deletion is that we don't need to check the point formation since we just verify the proof, right?

alxdavids: Yes that's right. We used to send all the proof values and check they were the same after recomputation. Then we verified the proof using the values that were sent by the server. This is a suboptimal approach and not as the protocol is described, exactly.

We now just verify the proof directly using only the (c,s) values as detailed in the protocol.

scripts/crypto.js
alxdavids: On the server-side all the points have this prefix. I'll make it static and add a comment explaining this though.

gtank: The byte 0x04 means "uncompressed" for SEC1 marshaling. 0x02 or 0x03 would indicate compression with a particular sign (as seen [here](https://github.com/privacypass/challenge-bypass-extension/blob/master/scripts/crypto.js#L214) and [here](https://github.com/privacypass/challenge-bypass-server/blob/master/crypto/curve.go#L65))

lziest: cool, assume this is the reference doc? http://www.secg.org/sec1-v2.pdf

alxdavids:
I've made the point encoding clearer in the latest commit. Has anyone had chance to check that the DLEQ verification steps match the protocol spec?

lziest:
@grittygrease I don't have privilege to merge this. :)

PR #39, opened 2018-10-22
New configurations for Privacy Pass to aid integrations

alxdavids:
* Workflow is now determine by config files. Eventually we hope to have configs for each different provider.
* Update signing to only send tokens when the cf-chl-bypass header is received.
* Also provide documentation to help consume the new config options and update the README.

This change is quite big because it involves refactoring a lot of the code so that we use the new variables in config.js. There is minimal new functionality beyond the new restriction on signing.

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -1,0 +1,127 @@
+# Privacy Pass configuration
+
+Privacy Pass interprets uses cryptographic tokens to bypass internet challenges from certain providers. In this doc, we discuss the configuration file that Privacy Pass uses for interpreting when tokens should be sent for signing/redemption.
+
+## config.js
+
+Holds the various configurations as JavaScript JSON structs for the providers that Privacy Pass interacts with. Currently, only the Cloudflare config is active. There is an example conifg, there is also an in progress config for the CAPTCHA provider FunCAPTCHA. In the following we will highlight how each config field is used.
+
+### config["id"]
+
+A unique identifier highlighting which config is used. Currently cfConfig.id = 1, all other configs should be added with unique "id" values.
+
+### config["sign"]
+
+A bool dictating whether Privacy Pass should send tokens for signing. 
+
+### config["redeem"]
+
+A bool dictating whether Privacy Pass should send tokens for redemption.
+
+### config["sign-reload"]
+
+A bool dictating whether the page should be reloaded after tokens are successfully signed
+
+### config["sign-resp-format"]
+
+Format of the response (as a string) to a signing request. Currently, support "string" or "json". When "string" is used, expect the signed tokens to be included (base-64 encoded) in the HTTP response body in the form `signatures= || <signed-tokens> || <Batch-DLEQ-Resp>`. When "json" is used, expect the signed tokens to be included (base-64 encoded) as a JSON struct with key: "signatures" and value `<signed-tokens> || <Batch-DLEQ-Resp>`.
+
+### config["storage-key-tokens"]
+
+The string key under which the tokens for the active config are stored in local storage. This should be unique for each config to prevent tokens being spent for inconsistent providers.
+
+### config["storage-key-count"]
+
+The string key corresponding to the number of tokens currently held in local storage (under config["storage-key-tokens"]). This should also be unique per config.
+
+### config["max-spends"]
+
+The integer number of tokens that should be redeemed per host in each interaction. This prevents Privacy Pass from repeatedly spending tokens for the same host (if some unknown issues occur, for example).
+
+### config["max-tokens"]
+
+The integer number of tokens that should be held by Privacy Pass at any one time.
+
+### config["tokens-per-request"]
+
+The integer number of tokens that should be sent with each signing request. For Cloudflare, there is also a server-side upper bound of 100 tokens for each signing request. We recommend that this is enforced to prevent unlimited numbers of tokens being signed.
+
+### config["var-reset"]
+
+A bool dictating whether the set of variables holding previous redemption information should be reset (see resetVars() in background.js). We recommend this is set to true, since these variables prevent tokens from being spent in the future.
+
+### config["var-reset-ms"]
+
+The time intervals (ms) by which the variables from above are reset.
+
+### config["commitments"]
+
+Hex-encoded elliptic curve commitment values for verifying DLEQ proofs. These essentially amount to public keys issued by the provider. The values of "G" and "H" held in config["commitments"]["dev"] should be used for development purposes. Those held in config["commitments"]["prod"] should be used in the production environment.
+
+### config["spending-restrictions"]
+
+A JSON struct of restrictions for redeeming tokens
+
+#### config["spending-restrictions"]["status-code"]
+
+An integer corresponding to the status code returned by the server. This HTTP status code is checked before a redemption is initiated, for Cloudflare this value is set to 403. That is, token redemptions can only occur after a HTTP response with status code 403 is received.
+
+#### config["spending-restrictions"]["iframe"]
+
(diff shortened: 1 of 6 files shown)


docs/CONFIG.md
```diff
@@ -0,0 +1,127 @@
+# Privacy Pass configuration
+
+Privacy Pass interprets uses cryptographic tokens to bypass internet challenges from certain providers. In this doc, we discuss the configuration file that Privacy Pass uses for interpreting when tokens should be sent for signing/redemption.
```
gabbifish: I would use "Privacy Pass uses..."

scripts/config.js
```diff
@@ -5,13 +5,113 @@
...
 /* exported ProdCommitmentConfig */
+/* exported CHL_BYPASS_SUPPORT */
+/* exported CHL_BYPASS_RESPONSE */
+/* exported PPConfigs */
 
-const DevCommitmentConfig = {
-    "G":"BIpWWWWFtDRODAHEzZlvjKyDwQAdh72mYKMAsGrtwsG7XmMxsy89gfiOFbX3RZ9Ik6jEYWyJB0TmnWNVeeZBt5Y=",
-    "H":"BKjGppSCZCsL08YlF4MJcml6YkCglMvr56WlUOFjn9hOKXNa0iB9t8OHXW7lARIfYO0CZE/t1SlPA1mXdi/Rcjo="
-}
+const CHL_BYPASS_SUPPORT  = "cf-chl-bypass"; // header from server to indicate that
```
gabbifish: Indicate what?

scripts/config.js
```diff
@@ -5,13 +5,113 @@
...
+const CHL_BYPASS_RESPONSE = "cf-chl-bypass-resp"; // response header from server, e.g. with erorr code
 
-const ProdCommitmentConfig = {
-    "G":"BOidEuO9HSJsMZYE/Pfc5D+0ELn0bqhjEef2O0u+KAw3fPMHHXtVlEBvYjE5I/ONf9SyTFSkH3mLNHkS06Du6hQ=",
-    "H":"BHOPNAWXRi4r/NEptOiLOp8MSwcX0vHrVDRXv16Jnowc1eXXo5xFFKIOI6mUp8k9/eca5VY07dBhAe8QfR/FSRY="
+const exampleConfig = {
```
gabbifish: This exampleConfig is super useful for a first-time user (it helped me familiarize myself with integrations)

scripts/background.js
```diff
@@ -145,85 +157,120 @@ function processHeaders(details) {
...
         }
 
-        // 403 with the right header indicates a bypassable CAPTCHA
-        if (isBypassHeader(header) && details.statusCode == 403) {
-            doRedeem = true;
+        // correct status code with the right header indicates a bypassable Cloudflare CAPTCHA
+        if (isBypassHeader(header) && SPEND_STATUS_CODE.indexOf(details.statusCode) > -1) {
+            let iframe = (details.frameId > 0);
+            // check if the token should only be spent on an iframe
+            if (SPEND_IFRAME) {
```
gabbifish:
```suggestion
            if ((SPEND_IFRAME && iframe) || !SPEND_IFRAME) {
                activated = true
            }
```
No need to set `activated` to false here--that's already its default.

scripts/background.js
```diff
@@ -145,85 +157,120 @@ function processHeaders(details) {
...
-            // Store the url for redirection after captcha is solved
-            // Manual check for favicon urls
-            storedUrl = url.href;
-            let faviconIndex = storedUrl.indexOf("favicon");
-            if (faviconIndex != -1) {
-                storedUrl = storedUrl.substring(0, faviconIndex);
+    if (activated && !spentUrl[url.href]) {
+        let redeemOccurred = false;
+        if (DO_REDEEM) {
+            if (countStoredTokens() > 0 && activated) {
```
gabbifish: Check for `activated` here seems to be already covered in line 177--I think this line can just be
```suggestion
            if (countStoredTokens() > 0) {
```

scripts/background.js
```diff
@@ -145,85 +157,120 @@ function processHeaders(details) {
...
+                redeemOccurred = attemptRedeem(url, details.tabId);
+            } else if (countStoredTokens() == 0) {
+                // Update icon to show user that token may be spent here
```
gabbifish: What happened to the logic for storing the url for redirection after the CAPTCHA is solved? Let me know if I'm missing something--was this logic moved elsewhere?

gabbifish:
Hi there! I'm still reviewing some of the code in this PR (you were right, it is big)!

Something that I think would be really good in the meantime is a testing suite that ensures this substantial refactor has not broken any functionality. I'll continue looking through the rest of this PR tomorrow morning :)

scripts/background.js
alxdavids: Yeah you're right, will fix.

scripts/background.js
alxdavids: True!

scripts/background.js
alxdavids: It appeared that storedUrl was never actually used beyond it being set. I just decided to remove it.

alxdavids:
I've pushed a new commit to the fork branch, but it seems like it has not been picked up here? I will check in a bit to see if it has updated.

scripts/background.js
```diff
@@ -145,85 +157,114 @@ function processHeaders(details) {
...
+            // If a clearance cookie is not held then set the spend flag
+            if (!clearanceHeld) {
+                fireRedeem(url, respTabId);
+                fired = true;
             }
         });
+    } else {
+        // If cookies aren't checked then we always attempt to redeem.
+        fireRedeem(url, respTabId);
+        return true;
```
gabbifish:
```suggestion
        fired = true
```
I think changing this line would make it clearer that a redemption has been fired

scripts/background.js
```diff
@@ -145,85 +157,114 @@ function processHeaders(details) {
...
+        let targetUrl = target[respTabId];
+        if (url.href == targetUrl) {
+            chrome.tabs.update(respTabId, { url: targetUrl });
+        } else if (!targetUrl || (targetUrl != url.href && !isFaviconUrl(targetUrl))) {
+            // set a reload in the future when the target has been inited
+            futureReload[respTabId] = url.href;
         }
-    });
+    } else {
+        throw new Error("[privacy-pass]: Incompatible redeem method selected.");
```
gabbifish: For now, I think specifying that reload is the only supported REDEEM_METHOD at the moment could be useful:
```suggestion
        throw new Error("[privacy-pass]: Incompatible redeem method selected. Only \"reload\" is supported at the moment.");
```

scripts/background.js
```diff
@@ -291,29 +372,28 @@ function beforeRequest(details) {
...
-    xhr.setRequestHeader("CF-Chl-Bypass", "1");
+    xhr.setRequestHeader(CHL_BYPASS_SUPPORT, "1");
     // We seem to get back some odd mime types that cause problems...
     xhr.overrideMimeType("text/plain");
 
-    xhr.send("blinded-tokens=" + request);
-
-    // Cancel the original request
-    return {redirectUrl: "javascript:void(0)"};
+    xhr.send(requestBody);
```
gabbifish: Why did we decide not to cancel the original request after sending the blinded tokens requestBody?
The code removed was 
```
    // Cancel the original request	
    return {redirectUrl: "javascript:void(0)"};
```

alxdavids: Good point. I replaced it with `return {cancel: true}` in the calling function, but the reason that we did it like this is to prevent a holding page appearing in Chrome. I've reinstated it now.

alxdavids:
I'm closing this PR as it has been succeeded by the PR in #54.

Issue #42, opened 2018-10-24
Testing framework


alxdavids:
We should use a testing framework for unit/functional testing to make building and testing the extension easier.

alxdavids:
This should be sorted now.

PR #44, opened 2018-10-31
Get ready for test framework

alxdavids:
For #42. 
We want to move towards a version of the extension where we can run integration tests through the code.
To do this, we need to set up yarn, node and jest. Have also moved the files relevant to the extension into addon/ and docs into docs/.
The actual tests are not part of this PR and will follow later.


henrydcase:
@alxdavids LGTM

PR #45, opened 2018-10-31
Small fixes

alxdavids:
- Fixes a bug where the correct spend flag was not set after tokens were received from an issue request (used url.href rather than url.host).
- Removes redundant use of storedUrl variable


henrydcase:
LGTM, I don't see any problems

PR #47, opened 2018-11-02
Integration tests for extension workflow

alxdavids:
Adds the actual integration tests for ensuring that the Privacy Pass workflow is running correctly. See tests/* for the new files.

Also change the boilerplates of some files to make them consistent with the rest. Added a couple of new methods in crypto.js that need to be exposed for some of the tests.

Final change for #42.


gabbifish:
Actually, one thing: it could be really cool to include test coverage! This isn't pressing, but is a really neat way to figure out how much every function is tested and if there's any functionality testing we may have missed. This is one tool that could be useful: https://istanbul.js.org/

This doesn't need to be part of this PR, this is just for future consideration :)

alxdavids:
Great! I agree about test coverage, I'll track it in a separate issue.

PR #53, opened 2018-11-07
Remove old ci file and add better dependency support

alxdavids:
* Moved dependencies from dev to production.
* Switched working directory to root
* Using yarn install rather than npm command
* New version of google-closure-compiler


alxdavids:
> Does the file addon/compiled/bg_compiled.js is autogenerated? If so, it must be excluded from the repo.
Otherwise, I considered is difficult to keep track on the changes of a minified source file.

Yes it does and good point. I've removed the minified source files from the repo in the latest commit.

> Is it a strong requirement that packages listed on package.json correspond to that specific version?

I guess not, but maybe having a fixed version is better as we are sure that it works in this state?

Issue #40, opened 2018-10-24
Use JSON configurations for workflow.


alxdavids:
We want to adopt a new workflow that works from a set configuration. This will make integrations with different providers and services much easier. It will also make for clearer documentation.

alxdavids:
This has been done now, closing.

PR #54, opened 2018-11-12
New configurations for Privacy Pass to aid integrations

alxdavids:
Changes for #40.

* Switch workflow to using separate configs, e.g. for each provider.
* Update signing to only send tokens when the cf-chl-bypass header is received.
* New tests and changes to some existing ones.
* Provide documentation (docs/CONFIG.md) to help consume the new config options and update the README.

Succeeds the PR in #39.

### DIFF
--- a/addon/scripts/background.js
+++ b/addon/scripts/background.js
@@ -19,26 +19,38 @@
 /* exported cookiesChanged */
 /* exported CHL_CAPTCHA_DOMAIN */
 /* exported CHL_CLEARANCE_COOKIE */
+/* exported REDEEM_METHOD */
 /* exported spentTab, spendId, timeSinceLastResp */
+/* exported ACTIVE_CONFIG */
 "use strict";
-
-const CHL_BYPASS_SUPPORT  = "cf-chl-bypass";
-const CHL_BYPASS_RESPONSE = "cf-chl-bypass-resp";
-const CHL_CLEARANCE_COOKIE = "cf_clearance";
-const CHL_CAPTCHA_DOMAIN = "captcha.website"; // cookies have dots prepended
-const CHL_VERIFICATION_ERROR = "6";
-const CHL_CONNECTION_ERROR = "5";
-const MAX_REDIRECT = 3;
-const SPEND_MAX = 3;
-const MAX_TOKENS = 300;
-const TOKENS_PER_REQUEST = 30;
-const FF_PRIV_TAB = "about:privatebrowsing";
-const CHROME_TABS = "chrome://";
-const FF_BLANK = "about:blank";
-const SERVER_REDIRECT = "server_redirect";
-const AUTO_SUBFRAME = "auto_subframe";
-const VALID_REDIRECTS = ["https://","https://www.","http://www."];
-const POTENTIALLY_GOOD_TRANSITIONS = ["link", "typed", "auto_bookmark", "reload"];
+/* Config variables that are reset in setConfig() depending on the header value that is received (see config.js) */
+let CONFIG_ID = ACTIVE_CONFIG["id"];
+let CHL_CLEARANCE_COOKIE = ACTIVE_CONFIG["cookies"]["clearance-cookie"];
+let CHL_CAPTCHA_DOMAIN = ACTIVE_CONFIG["captcha-domain"]; // cookies have dots prepended
+let CHL_VERIFICATION_ERROR = ACTIVE_CONFIG["error-codes"]["connection-error"];
+let CHL_CONNECTION_ERROR = ACTIVE_CONFIG["error-codes"]["verify-error"];
+let SPEND_MAX = ACTIVE_CONFIG["max-spends"];
+let MAX_TOKENS = ACTIVE_CONFIG["max-tokens"];
+let DO_SIGN = ACTIVE_CONFIG["sign"];
+let DO_REDEEM = ACTIVE_CONFIG["redeem"];
+let RELOAD_ON_SIGN = ACTIVE_CONFIG["sign-reload"];
+let SIGN_RESPONSE_FMT = ACTIVE_CONFIG["sign-resp-format"];
+let REDEEM_METHOD = ACTIVE_CONFIG["spend-action"]["redeem-method"];
+let HEADER_NAME = ACTIVE_CONFIG["spend-action"]["header-name"];
+let TOKENS_PER_REQUEST = ACTIVE_CONFIG["tokens-per-request"];
+let SPEND_STATUS_CODE = ACTIVE_CONFIG["spending-restrictions"]["status-code"];
+let MAX_REDIRECT = ACTIVE_CONFIG["spending-restrictions"]["max-redirects"];
+let NEW_TABS = ACTIVE_CONFIG["spending-restrictions"]["new-tabs"];
+let BAD_NAV = ACTIVE_CONFIG["spending-restrictions"]["bad-navigation"];
+let BAD_TRANSITION = ACTIVE_CONFIG["spending-restrictions"]["bad-transition"];
+let VALID_REDIRECTS = ACTIVE_CONFIG["spending-restrictions"]["valid-redirects"];
+let VALID_TRANSITIONS = ACTIVE_CONFIG["spending-restrictions"]["valid-transitions"];
+let VAR_RESET = ACTIVE_CONFIG["var-reset"];
+let VAR_RESET_MS = ACTIVE_CONFIG["var-reset-ms"];
+const STORAGE_STR = "bypass-tokens-";
+const COUNT_STR = STORAGE_STR + "count-";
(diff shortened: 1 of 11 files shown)


test/headersReceived.test.js
```diff
@@ -109,7 +114,110 @@ describe("check redemption attempt conditions", () => {
...
+                url = new URL("https://captcha.website");
+                getMock = function() { return 2; };
+                workflow.__set__("get", getMock);
+                let fired = processHeaders(details, url);
+                expect(fired).toBeFalsy();
+                let readySign = workflow.__get__("readySign");
+                expect(readySign).toBeTruthy();
+            });
+
+            test("redemption off", () => {
```
gabbifish: Maybe add one more unit test for background.js:168 (that is, checking that the button icon becomes (!) when token count == 0

addon/scripts/background.js
```diff
@@ -140,27 +151,29 @@ function processHeaders(details, url) {
...
     let attempted = false;
-    if (doRedeem && !spentUrl[url.href]) {
-        if (countStoredTokens() > 0) {
-            // Prevent reloading on captcha.website
-            if (url.host.indexOf(CHL_CAPTCHA_DOMAIN) != -1) {
-                return;
+    if (activated && !spentUrl[url.href]) {
+        let count = countStoredTokens();
+        if (DO_REDEEM) {
+            if (count > 0 && activated && url.host.indexOf(CHL_CAPTCHA_DOMAIN) == -1) {
```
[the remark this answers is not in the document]
alxdavids: Done.

test/headersReceived.test.js
alxdavids: Good call, I've added an extra line to the test "no tokens" (headersReceived.test.js:167) that checks whether the updateIcon method was called with "!".

addon/scripts/background.js
```diff
@@ -480,16 +556,26 @@ function badTransition(href, type, transitionType) {
...
     if (!type && !maybeGood) {
         return true;
     }
-    return type == SERVER_REDIRECT;
+    return BAD_TRANSITION.indexOf(type) > -1;
+}
+
+// Mark the url so that a sign doesn't occur again.
+function markSignUrl(url) {
+    return url + "&captcha-bypass=true";
```
armfazh: Is it possible to hand-made construct an URL with `&captcha-bypass=false` ? are there any side effects doing that?

test/beforeHeadersSend.test.js
```diff
@@ -97,14 +115,16 @@ describe("redemption attempted", () => {
...
         let reqHeaders = redeemHdrs.requestHeaders;
-        expect(reqHeaders).toBeTruthy();
-        expect(reqHeaders[0].name == "challenge-bypass-token").toBeTruthy();
-        expect(reqHeaders[0].value == b64EncodedToken).toBeTruthy();
+        expect(getSpendFlag(url.host)).toBeFalsy();
         expect(getSpendId([details.requestId])).toBeTruthy();
         expect(getSpentUrl([url.href])).toBeTruthy();
         expect(getSpentTab([details.tabId]) == url.href).toBeTruthy();
+        expect(reqHeaders).toBeTruthy();
+        expect(reqHeaders[0].name == "challenge-bypass-token").toBeTruthy();
```
armfazh: Does this string `"challenge-bypass-token"` must be a constant ? or it is safe to use the literal value?

docs/CONFIG.md
```diff
@@ -0,0 +1,123 @@
+# Privacy Pass configuration
+
+Privacy Pass uses cryptographic tokens to bypass internet challenges from certain providers. In this doc, we discuss the configuration file that Privacy Pass uses for interpreting when tokens should be sent for signing/redemption. See config.js for an example configuration (exampleConfig) describing the correct format of the JSON struct.
+
+## config.js
+
+Holds the various configurations as JavaScript JSON structs for the providers that Privacy Pass interacts with. Currently, only the Cloudflare config is active. There is an example conifg, there is also an in progress config for the CAPTCHA provider FunCAPTCHA. In the following we will highlight how each config field is used.
```
armfazh: conifg -> config

addon/scripts/browserUtils.js
```diff
@@ -22,42 +22,59 @@
...
+    // Check all cookie stores to see if a clearance cookie is held
+    if (CHECK_COOKIES) {
+        chrome.cookies.getAllCookieStores(function(stores) {
+            let clearanceHeld = false;
+            stores.forEach( function(store, index) {
+                var tabIds = store.tabIds;
+                if (tabIds.length && tabIds[0].id !== undefined) {
+                    tabIds = tabIds.map((tab) => respTabId.id);
+                }
+                var storeMatches = tabIds.indexOf(respTabId) > -1;
```
henrydcase: I don't see a reason for storeMatches to exist

docs/CONFIG.md
```diff
@@ -0,0 +1,123 @@
...
+
+### config["sign-reload"]
+
+A bool dictating whether the page should be reloaded after tokens are successfully signed
+
+### config["sign-resp-format"]
+
+Format of the response (as a string) to a signing request. Currently, support "string" or "json". When "string" is used, expect the signed tokens to be included (base-64 encoded) in the HTTP response body in the form `signatures= || <signed-tokens> || <Batch-DLEQ-Resp>`. When "json" is used, expect the signed tokens to be included (base-64 encoded) as a JSON struct with key: "signatures" and value `<signed-tokens> || <Batch-DLEQ-Resp>`.
+
+### config["storage-key-tokens"]
```
henrydcase: I would like to challenge a fact that this is really needed. Why not to simply have a directory that's named after config ID? This would remove a needed for "storage-key-tokens" to be a unique name.

docs/CONFIG.md
```diff
@@ -0,0 +1,123 @@
...
+
+The string key under which the tokens for the active config are stored in local storage. This should be unique for each config to prevent tokens being spent for inconsistent providers.
+
+### config["storage-key-count"]
+
+The string key corresponding to the number of tokens currently held in local storage (under config["storage-key-tokens"]). This should also be unique per config.
```
henrydcase: "This should also be unique per config." 
Why?

addon/scripts/background.js
```diff
@@ -211,30 +228,59 @@ function beforeSendHeaders(request, url) {
...
     }
 
+    // Only sign tokens if config says so and the appropriate header was received previously
+    if (!DO_SIGN || !readySign) {
+        return false;
+    }
+
+    // Different signing methods based on configs
+    let xhrInfo;
+    switch (CONFIG_ID) {
```
henrydcase: Why not "if CONFIG_ID != 1 { throw() }  doStuf() " ?

addon/scripts/background.js
alxdavids: Yeah there are no side-effects to doing this, it would just cause the signing operation to happen again. The same as if the URL was intentionally stripped of the query string.

addon/scripts/background.js
```diff
@@ -140,27 +151,29 @@ function processHeaders(details, url) {
...
     let attempted = false;
-    if (doRedeem && !spentUrl[url.href]) {
-        if (countStoredTokens() > 0) {
-            // Prevent reloading on captcha.website
-            if (url.host.indexOf(CHL_CAPTCHA_DOMAIN) != -1) {
-                return;
+    if (activated && !spentUrl[url.href]) {
+        let count = countStoredTokens();
+        if (DO_REDEEM) {
+            if (count > 0 && url.host.indexOf(CHL_CAPTCHA_DOMAIN) == -1) {
```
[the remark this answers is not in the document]
alxdavids: Will standardise using tabIds.length > 0

addon/scripts/background.js
alxdavids: If we want to introduce new configurations, then maybe the switch is better?

alxdavids:
@henrydcase @armfazh I've committed changes based on your comments. Let me know if you think they are addressed sufficiently.

@armfazh In answer to your questions:
- according to https://yarnpkg.com/lang/en/docs/yarn-lock/#toc-managed-by-yarn: 

> All yarn.lock files should be checked into source control

so yes, it should be included.

- regarding sjcl.js, that's a good point regarding the version and whether we should retrieve it from source. According to https://bitwiseshiftleft.github.io/sjcl/ we should use a local version for using EC operations:

> SJCL also supports elliptic curve cryptography (ECDH public key encryption and ECDSA signatures). If you want to use it, you will have to compile your own version of sjcl.js

However, I will try to find out which version we are using and add it in a later commit for clarity.

docs/CONFIG.md
alxdavids: Removed. Using the same id method as the previous comment.

addon/scripts/background.js
henrydcase: personally I avoid adding things that complicate stuff but "may be useful later". but here that's just small details. All good - LGTM

Issue #50, opened 2018-11-05
Null point errors


alxdavids:
An error that occurs occasionally is that the blind point is null. There is currently no check that ascertains whether this point is null and fails more gracefully (or even better, retries).

alxdavids:
This was completed in #55

PR #55, opened 2018-11-19
Prevent null point errors

alxdavids:
Sometimes the random token generation would fail and curve points would be null. This would cause uncaught errors that would stop token generation.

Rewritten the token geenration logic so that null curve points are caught and ignored.

This addresses #50.

### DIFF
--- a/addon/scripts/tokens.js
+++ b/addon/scripts/tokens.js
@@ -23,17 +23,27 @@
 //  r blinding factor, sjcl bignum
 function CreateBlindToken() {
     let t = newRandomPoint();
-    let bpt = blindPoint(t.point);
-    return { token: t.token, point: bpt.point, blind: bpt.blind };
+    let tok;
+    if (t) {
+        let bpt = blindPoint(t.point);
+        tok = { token: t.token, point: bpt.point, blind: bpt.blind };
+    }
+    return tok;
 }
 
 // returns: array of blind tokens
 function GenerateNewTokens(n) {
-    let i = 0;
-    let tokens = new Array(n);
-    for (i = 0; i < tokens.length; i++) {
-        tokens[i] = CreateBlindToken();
+    let tokens = [];
+    for (let i=0; i<n; i++) {
+        let tok = CreateBlindToken();
+        if (!tok) {
+            console.warn("[privacy-pass]: Tried to generate a random point on the curve, but failed.");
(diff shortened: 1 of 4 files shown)


addon/scripts/tokens.js
```diff
@@ -23,16 +23,27 @@
 //  r blinding factor, sjcl bignum
 function CreateBlindToken() {
     let t = newRandomPoint();
```
henrydcase: I think I would either return value that's always correctly initialize or ``null``. makes error handling easier

addon/scripts/tokens.js
```diff
@@ -23,16 +23,27 @@
...
 }
 
 // returns: array of blind tokens
 function GenerateNewTokens(n) {
+    let tokens = [];
     let i = 0;
-    let tokens = new Array(n);
-    for (i = 0; i < tokens.length; i++) {
-        tokens[i] = CreateBlindToken();
+    while (i < n) {
```
henrydcase: seems to me using `` for (var i=0; i<n; i++)`` makes more sense here

addon/scripts/tokens.js
```diff
@@ -23,16 +23,27 @@
...
+        let tok = CreateBlindToken();
+        if (!tok) {
+            console.warn("[privacy-pass]: Tried to generate a random point on the curve, but failed.");
+            console.warn("[privacy-pass]: Will drop the null point.");
+            n--;
```
henrydcase: this looks weird (as far as my javascript knowledge goes). as n is passed by copy and is not read anywhere in this function, I think reducing it has no effect. Especially that n is a lvalue.
see https://js.do/code/axrev1

addon/scripts/tokens.js
alxdavids: yes, good idea.

addon/scripts/tokens.js
alxdavids: You're right this is a weird way of doing it. I wrote it this way to prevent empty array entries. However I've rewritten it so that the array is populated and then empty entries are removed afterwards.

addon/scripts/tokens.js
alxdavids: (see above)

alxdavids:
@henrydcase If you're okay with my fixes for your comments then I will merge.

Issue #56, opened 2018-11-23
Support key rotation


alxdavids:
We need to be able to change the server's key easily. This requires changing the commitments that the extension uses to verify DLEQ proofs. The functionality should be:
* Server sends version parameter (either with tokens, as an additional HTTP header or in html of webpage)
* Extension loads commitments corresponding to version
* When tokens are signed, the extension verifies the DLEQ proof with the loaded commitments

To make the key rotation procedure easier, we should load the commitments from a separate github respository (privacypass/ec-commitments). In the future, this may also make it easier to support multiple EC configurations.

alxdavids:
We now read the commitments from github.com/privacypass/ec-commitments. Currently we use tokens that correspond to v1.0 of the key. Still need to decide exactly how rotation will work in terms of the server notifying which commitments the client has to use.

alxdavids:
Key rotation is now handled by specifying a "version" that determines which commitments are used for verifying the proof. This version is sent with tokens that have been signed by the server.

PR #57, opened 2018-11-23
Read P256 commitments from a separate repository

alxdavids:
To make key rotation easier in the future, we want to read the public EC commitiments from a separate file (#56). 
* The keys are organised by version.
* They are retrieved via an XHR request when the active configuration is determined (url: https://github.com/privacypass/ec-commitments).
* In the future, the server will respond with a version parameter that determines which commitment to read.
*We only support version "1.0" for now, will support actual versioning in a future PR.

There is also a minor change to make the futureReload variable global as this is consistent with our other conventions. I also observed some weird behaviour with redemptions that required me to reset the variable reset timer when a redemption is activated.


alxdavids:
@henrydcase I've updated, let me know if you have any issues.

PR #58, opened 2018-11-30
Allow specifying commitment version from issuance responses

alxdavids:
This change finishes the work required for allowing key rotation (#56) via versioning. By versioning, we mean that server responses to signing requests can now contain a version parameter that specifies which commitments to use for verifying the DLEQ proof.

To make this change possible, we are now moving towards accepting issuance responses of the form:
```
{
	sigs: <signed-tokens>,
	proof: <batch-proof>,
	version: <version-string>
}
```
see the README changes for more details.

Other major changes include reading the commitments when the signed tokens are received, rather than when the configuration is set. This means effectively running the token storage and proof verification logic as a callback for when the commitments are received via XHR.

Also refactored some functions to more appropriate locations.


addon/scripts/crypto.js
```diff
@@ -281,6 +278,100 @@ function decodeStorablePoint(s) {
...
+    let commH;
+    const respBody = xhr.responseText;
+    let resp = JSON.parse(respBody);
+    let comms = resp[COMMITMENTS_KEY];
+    if (comms) {
+        if (DEV) {
+            commG = comms["dev"]["G"];
+            commH = comms["dev"]["H"];
+        } else if (version) {
+            commG = comms[version]["G"];
```
armfazh: Q: Does version can be `"dev"` string? Is there a checking/validation process for the version parameter?

alxdavids: Yes it could be dev. Good point about the validation. I've put a null check outside of the function to check that the commitments are actually defined after retrieval.

addon/scripts/background.js
```diff
@@ -398,6 +380,23 @@ function parseSigString(data) {
     return atob(split[1]);
 }
 
+// Retrieves the batchProof and signatures, depending on the type of object
+// received
+function parsePointsAndProof(issueResp) {
+    let signatures;
+    let batchProof;
+    let version;
+    if (!issueResp[0]) {
```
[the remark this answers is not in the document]
alxdavids: Good point, I've added a comment. It essentially just checks if issueResp is an array or not. The old method used an array, but I think it's going to be easier if we move to a JSON struct on the server side.

alxdavids:
@armfazh Thanks for your questions.

1. The version is only necessary for the verification of the DLEQ proof. Therefore the format of the responses do not change. The server will just attempt to validate the tokens using the currently available keys.
2. Good point. I intend that for each version it will just be a key rotation server-side but I will make this clear in the future.
3. It depends on the key rotations that have occurred on the server-side. The server currently signs tokens using one key, and then attempts to validate tokens using the same key, and a selection of old keys. If the tokens are really old the verification will just fail. The errors are sent in a header and the extension witnesses them. You have just reminded me that we should clear the tokens out if such an error occurs so that the client can receive new tokens. I will do this in a future change.

Issue #26, opened 2018-03-15
Hash to curve optimisations


alxdavids:
https://github.com/privacypass/challenge-bypass-extension/blob/master/scripts/crypto.js#L220
It might be quicker to use the Jacobian rather than the square root for checking if the point is on the curve.

alxdavids:
Another optimisation we should consider is moving from try-and-increment to optimal SWU as described in https://eprint.iacr.org/2009/340.pdf.

alxdavids:
Added an optimised SWU implementation for hashing to curve in #60. We will still use the hash-and-increment method until we can make a change in the server to be consistent.

PR #60, opened 2018-12-12
Add SWU method for hashing bytes to P256

alxdavids:
Implements the Shallue-Woestijne-Ulas (SWU) algorithm for hashing bytes to points on elliptic curves (e.g. P256) for #26. We turn this setting off by default since it needs to be compatible with the server (where SWU is not currently implemented).

In the future we want to deprecate the old hash-and-increment algorithm that is being used.

The code is not currently constant-time (although this has always been the case previously, anyway). The lack of constant-time code is unlikely to be a problem for Privacy Pass as the hash-to-curve algorithm is only used after the user is queried for a CAPTCHA. Happy to hear more input on this.


addon/scripts/h2c.js
```diff
@@ -0,0 +1,204 @@
...
+
+  // step 4
+  const expo = p.add(1).cnormalize().halveM().halveM();
+  let Y = new FFp(g.power(expo));
+
+  // step 5-8
+  const Ysq = Y.square();
+  const b = g.equals(Ysq);
+  const cX = c.mul(X);
+  const t3Y = t.power(3).mul(Y); // p+1/4 is even for P256
```
wbl: Why is this the correct y value? It isn't clear to me from reading the RFC. If there is some algebra involved in that, it would be nice to have an explanation to make the correctness clear.

addon/scripts/h2c.js
```diff
@@ -0,0 +1,204 @@
...
+  X = b ? X : cX;
+  Y = b ? Y : t3Y;
+
+  // step 9
+  return new sjcl.ecc.point(p256, X, Y);
+}
+
+/**
+ * Converts field elements into jacobian coordinates on P256
```
gabbifish: Could be useful to link to document that describes this conversion

alxdavids:
@gabbifish Thanks for your comments. I agree that the implementation should probably match the RFC exactly. I am currently implementing an optimised form of the algorithm (from @armfazh) and it might be better If I have this recognised in a formal change to the document before I make changes here.

addon/scripts/h2c.js
```diff
@@ -0,0 +1,204 @@
...
+ * @param {sjcl.bn} B (see above)
+ * @param {sjcl.bn} t FF_p field element
+ * @return {sjcl.ecc.pointJac} Jacobian coordinates of a curve point
+ */
+function jacobianSWUP256(p, A, B, t) {
+  // calculate X/Z
+  let u = t.square();
+  u = u.mul(-1);
+  let Z = new FFp(u.square().add(u));
+  let X = new FFp(B.mul(Z.add(1)));
```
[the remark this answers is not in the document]
alxdavids: This is the jacobian version of the algorithm, which I believe isn't part of the draft yet? @armfazh what do you think?

addon/.eslintrc.json
```diff
@@ -74,7 +74,9 @@
         "loadTokens": true,
         "countStoredTokens": true,
         "STORAGE_KEY_TOKENS": true,
-        "STORAGE_KEY_COUNT": true
+        "STORAGE_KEY_COUNT": true,
+        "HASH_AND_INC": true,
+        "h2Curve": true
```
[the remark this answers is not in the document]
alxdavids: I've switched to using a string (enums appear to not be a formal concept in JS) for the hash-to-curve that is used in config.js.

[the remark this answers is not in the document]
alxdavids: I've switched to using a string (enums appear to not be a formal concept in JS) for the hash-to-curve that is used in config.js.

addon/scripts/h2c.js
```diff
@@ -0,0 +1,207 @@
...
+function affineSWUP256(p, A, B, t) {
+  // step 1
+  let u = t.square();
+  u = u.mul(-1);
+
+  // step 2
+  let t0 = u.square().add(u);
+  t0 = t0.inverseMod(p);
+  t0 = t0.add(1);
+  const invA = A.inverseMod(p);
```
gabbifish: Perhaps add Armando's documentation as a markdown document along with this repo, so people can follow the explanation there. I'd also renumber the steps for simplified SWU so they are consistent with Armando's documentation.

addon/scripts/h2c.js
```diff
@@ -0,0 +1,207 @@
...
+  let X = new FFp(B.mul(invA).mul(-1).mul(t0));
```
gabbifish: It looks like this is missing a -B/A term? (step 4 of affine SWU)

alxdavids: It may not be clear, but this is what:
`B.mul(invA).mul(-1)`
is doing. I'll put it on a new line to emphasise what is happening.

addon/scripts/h2c.js
alxdavids: Good idea, I will do that.

alxdavids:
(rebased onto new master)

alxdavids:
Rebased onto master and squashed all the commits into one. Should be ready to check again.

Issue #65, opened 2019-01-30
Hash to curve configuration to be sent to server


alxdavids:
For each token that is signed by the server, the client should store the configuration that was used to create the token. This configuration includes the curve (P256 currently), the hash function that is used to hash to the base field and the actual hash-to-curve algorithm that is used.

This should be stored as a JSON struct of the form:

```
{
  "curve": "p256",
  "base-hash": "sha256",
  "hash-to-curve": "swu"
}
```
for example. When the token is then chosen for redemption, this JSON struct should be sent to the server so that the server knows which configuration to use when validating the token. 

This would make it easier in the future for the client & server to swap between the various algorithms that are used and may make switching to new elliptic curves in the future.

alxdavids:
Maybe we should use `"method"` instead of `"hash-to-curve"`

alxdavids:
The contents of `h2c-params` are sent by the client if the value of `send-h2c-params` is true. Otherwise they are not sent. The reason we do this is because the data format that is used is suboptimal right now (#73) and so we should fix that before we add on more data to redemption requests.

PR #71, opened 2019-02-06
Send hash-to-curve parameters during token redemptions

alxdavids:
For #65.
* Hash-to-curve parameters are stored in client configurations
* Configurations determine which curve setting is used when client is performing cryptographic operations
* The curve configuration is sent from the client to the server when tokens are redeemed.
* also: 
  - moved hashAndInc into h2c.js and have it controlled by the h2Curve function
  - changed token.token to token.data to avoid naming collisions 
  - fixed the linter to actually include the unminified src


src/ext/h2c.js
```diff
@@ -2,15 +2,17 @@
...
  */
 
  /* global sjcl */
  /* exported h2Curve */
 
-// compatible curves
-const p256 = sjcl.ecc.curves.c256;
-const FFp = p256.field;
+ // We only support hashing to curve using the affine representation for now
+ const SWU_MODE = 0;
```
armfazh: Here is good to indicate the meaning of `SWU_MODE = 0` and other possible options

src/ext/h2c.js
```diff
@@ -34,43 +34,43 @@ function h2Base(x, curve, label) {
...
+  switch (ecSettings.method) {
+    case "swu":
+      point = simplifiedSWU(alpha, ecSettings.curve, ecSettings.hash, SWU_MODE);
+      break;
+    case "increment":
+      point = hashAndInc(alpha, ecSettings.curve, ecSettings.hash);
       break;
     default:
-      throw new Error("[privacy-pass]: Incompatible curve chosen for hashing: " + curveName);
+      throw new Error("[privacy-pass]: Incompatible curve chosen for hashing: " + sjcl.ecc.curveName(ecSettings.curve));
```
armfazh: This line will show `c256` instead of `p256`

test/beforeHeadersSend.test.js
```diff
@@ -16,7 +17,7 @@ var URL = window.URL;
 const EXAMPLE_HREF = "https://www.example.com";
 const beforeSendHeaders = workflow.__get__('beforeSendHeaders');
 const setConfig = workflow.__get__('setConfig');
-const b64EncodedToken = "eyJ0eXBlIjoiUmVkZWVtIiwiY29udGVudHMiOltbMjQsNjIsNTYsMTAyLDc2LDEyNywyMDEsMTExLDE2MSwyMTgsMjQ5LDEwOSwzNCwxMjIsMTYwLDIxOSw5MywxODYsMjQ2LDEyLDE3OCwyNDksMjQxLDEwOCw2OSwxODEsNzcsMTQwLDE1OCwxMywyMTYsMTg0XSxbMjI3LDExLDk1LDIxNSwxNSwyMTUsMTM1LDI0LDEzNywxNzQsMjMzLDgsODYsMTQ4LDEzMCwxOTEsNDYsMTgzLDkyLDEwOCwxNjAsMjQ5LDE1OCwyMzEsMTU5LDIxOCwyNTQsODAsMTQ4LDQ0LDI5LDI1NF1dfQ==";
+const b64EncodedToken = "eyJ0eXBlIjoiUmVkZWVtIiwiY29udGVudHMiOltbMjQsNjIsNTYsMTAyLDc2LDEyNywyMDEsMTExLDE2MSwyMTgsMjQ5LDEwOSwzNCwxMjIsMTYwLDIxOSw5MywxODYsMjQ2LDEyLDE3OCwyNDksMjQxLDEwOCw2OSwxODEsNzcsMTQwLDE1OCwxMywyMTYsMTg0XSxbMjI3LDExLDk1LDIxNSwxNSwyMTUsMTM1LDI0LDEzNywxNzQsMjMzLDgsODYsMTQ4LDEzMCwxOTEsNDYsMTgzLDkyLDEwOCwxNjAsMjQ5LDE1OCwyMzEsMTU5LDIxOCwyNTQsODAsMTQ4LDQ0LDI5LDI1NF0sWzEyMywzNCw5OSwxMTcsMTE0LDExOCwxMDEsMzQsNTgsMzQsMTEyLDUwLDUzLDU0LDM0LDQ0LDM0LDEwNCw5NywxMTUsMTA0LDM0LDU4LDM0LDExNSwxMDQsOTcsNTAsNTMsNTQsMzQsNDQsMzQsMTA5LDEwMSwxMTYsMTA0LDExMSwxMDAsMzQsNTgsMzQsMTA1LDExMCw5OSwxMTQsMTAxLDEwOSwxMDEsMTEwLDExNiwzNCwxMjVdXX0=";
```
armfazh: Size of tokens has increased significantly, this could be an overhead in some scenarios. Any thoughts about token's size?

alxdavids: @armfazh That is true, the size increases by over 200 bytes because I the base-64 encoding is called over the string representation of the byte array of the curve settings (which is only 53 bytes in fact).

We could significantly reduce this by encoding each of the byte arrays into hexidecimal before calling btoa() on it. I think this is worthwhile doing but it will also require a change to the data format that is sent by the client, so maybe it would be better suited to a separate change?

src/ext/h2c.js
alxdavids: I think this is unavoidable without rewriting the naming function, I will add "SJCL curve: " to the string to make it clear what we mean.

test/beforeHeadersSend.test.js
armfazh: I agree, this is a separate issue.

Issue #73, opened 2019-02-11
Better data format


alxdavids:
Data that is sent in the redemption step of the protocol is encoded into byte arrays and then the byte arrays are encoded into strings directly rather then performing a better encoding. This results in higher packet sizes than is necessary so we should fix that.

A new format may not be compatible with the server-side so we will need to ensure a smooth rollover.

alxdavids:
This has been fixed, the individual array elements in redemption requests are now base64 encoded.

PR #74, opened 2019-02-11
Better data format for redemptions

alxdavids:
For #73. Redemption data is now an array of base64-encoded values, rather than ArrayBuffers interpreted directly as a string. This reduces the amount of data that we need to send to the server significantly.


alxdavids:
Rebased on master and removed an old comment, now merging.

PR #77, opened 2019-02-25
Local fixes for curve implementation

alxdavids:
- Use active curve setting rather than hardcoded for local crypto operations
- Point decompression now works

### DIFF
--- a/src/ext/h2c.js
+++ b/src/ext/h2c.js
@@ -42,7 +42,7 @@ function h2Curve(alpha, ecSettings) {
         point = simplifiedSWU(alpha, ecSettings.curve, ecSettings.hash, SWU_POINT_REPRESENTATION);
         break;
     case "increment":
-        point = hashAndInc(alpha, ecSettings.curve, ecSettings.hash);
+        point = hashAndInc(alpha, ecSettings.hash);
         break;
     default:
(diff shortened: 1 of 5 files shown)


alxdavids:
@armfazh Fixed so that the tag is checked. I've also updated all the comments in the file and also added some fixes for hash-and-inc which were not updated yet. Let me know if this is all good now.

src/ext/h2c.js
```diff
@@ -233,13 +232,16 @@ function hashAndInc(seed, curve, hash) {
...
-        let point = decompressPoint(digestBits, curve, 0x02);
+        // attempt to decompress a point along with valid tags
+        // curve choice is implicit based on active curve parameters
+        let point = decompressPoint([2].concat(bytes));
         if (point !== null) {
             return point;
         }
 
-        point = decompressPoint(digestBits, curve, 0x03);
+        point = decompressPoint([3].concat(bytes));
```
armfazh: #### I noticed a logical error on this method.

Observe that trying to decode a point using `P = decode( 0x02 || bytes) ` is essentially the same as decoding `P = decode( 0x03 || bytes) `. Both points have the same `x` coordinate but different `y` coordinate. 
Then, if the first call succeeds, then the point in the second call is just the negative point obtained in the first call.
On the other hand, if the first call fails on decoding a point, the second call will fail as well. Hence it spent twice the time on every failure.

Issue #78, opened 2019-02-26
double-time spent on hash&increment


armfazh:
#### Hash&Increment

On failure, the decoding is taking double of the time since it tries to decode a point with the same `x` coordinate. 

Observe that trying to decode a point using `P = decode( 0x02 || bytes) ` is essentially the same as decoding `P = decode( 0x03 || bytes) `. Both points have the same `x` coordinate but different `y` coordinate. Then, if the first call succeeds, then the point in the second call is just the negative point obtained in the first call.
On the other hand, if the first call fails on decoding a point, the second call will fail as well. Hence it spent twice the time on every failure.

alxdavids:
Thanks for pointing this out, so we just need to remove the second decoding step.

armfazh:
yes, it seems to be the simplest way to fix it

alxdavids:
This has now been fixed.

PR #79, opened 2019-03-04
Optimisations for cryptographic operations and set cfConfig to default

alxdavids:
For #78. 
* Fixes h2c to only test points with the tag 0x02 (since 0x03 is just negative version of same point). 
* For computing composites, apply all operations in jacobian format and switch to affine at the end.
* Set active config to cfConfig as default (rather than using exampleConfig as default).
* Removed trailing spaces in beforeRequest.test.js

### DIFF
--- a/src/crypto/local.js
+++ b/src/crypto/local.js
@@ -313,22 +313,17 @@ function recomputeComposites(chkM, chkZ, pointG, pointH) {
     let seed = getSeedPRNG(chkM, chkZ, pointG, pointH);
     let shake = createShake256();
     shake.update(seed, "hex");
-    let cM;
-    let cZ;
+    let cM = new sjcl.ecc.pointJac(CURVE); // can only add points in jacobian representation
+    let cZ = new sjcl.ecc.pointJac(CURVE);
     for (let i=0; i<chkM.length; i++) {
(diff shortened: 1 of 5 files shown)


src/crypto/local.js
```diff
@@ -313,22 +313,17 @@ function recomputeComposites(chkM, chkZ, pointG, pointH) {
...
-        } else {
-            cM = cM.toJac().add(cMi).toAffine();
-            cZ = cZ.toJac().add(cZi).toAffine();
-        }
+        cM = cM.add(cMi);
+        cZ = cZ.add(cZi);
     }
 
-    return {M: cM, Z: cZ};
+    return {M: cM.toAffine(), Z: cZ.toAffine()};
```
armfazh: You can delay the conversion to affine, because the caller function uses these point to perform another scalar multiplication.

armfazh: Another related optimization that works is replacing two consecutive scalar mults by calling the method `mult2` (which calculates `aP+bQ`) provided by sjcl.


For example in lines (275-276) and (280-281)

alxdavids: Good point, I'll see if I can add this in a future change

PR #80, opened 2019-03-05
Better retrieval from cache

alxdavids:


test/beforeRequest.test.js
```diff
@@ -146,9 +146,6 @@ beforeEach(() => {
     setMockFunctions();
     setTimeSinceLastResp(Date.now());
     setConfig(1); // set the CF config
-    // Mock the active commitments because XHR is not available
```
henrydcase: Is removing a comment intendent?

alxdavids: Yeah, this was an old leftover from when commitments were stored directly in the extension.

PR #81, opened 2019-03-08
hCaptcha support

durch:
This PR adds support for the [hCaptcha](https://hcaptcha.com) challenge. 

The biggest single change is the introduction of a `no-reload` redemption method. As our captcha is served asynchronously, embedded in the source page, the `reload` paradigm did not fit. 

All our captchas are served from the same set of endpoints, all the verification requests also go to same, but different, set of endpoints every time. This is reflected in fixed URL patterns in `hcConfig.spend-action.urls`. We also introduce a new configuration key, `issue-action`, to handle issue specific URL patterns. 

In addition we introduce new keys in the `spend-action` key, `header-host-name` and `header-path-name`. These are quite specific, and we're open to dropping them from the config in case they are deemed too specific.

There are also quite a few obsessive style changes, such as `===`, or trailing `,`. While we'd prefer to at least keep the strict (in)equality checks, we're open to dropping those.

We are also very open to feedback about our issue and redemption approach, any and all suggestions or critiques are very welcome.

Cheers!


src/ext/background.js
```diff
@@ -207,10 +209,45 @@ function processHeaders(details, url) {
...
+            headers.push({name: ACTIVE_CONFIG["spend-action"]["header-host-name"], value: url.hostname});
+            headers.push({name: ACTIVE_CONFIG["spend-action"]["header-path-name"], value: http_path});
+            spendId[request.requestId] = true;
+            spentUrl[reqUrl] = true;
+            return {requestHeaders: headers};
+        }
+    }
+
+    if (!DO_REDEEM || !getSpendFlag(url.host) || checkMaxSpend(url.host) || spentUrl[reqUrl] || isErrorPage(reqUrl) || isFaviconUrl(url.href) || REDEEM_METHOD !== "reload") {
+        // TODO reverse this branch, end up with returning {cancel: false}
```
alxdavids: Agree, we should return {cancel:false} at the end of the function.

src/ext/issuance.js
```diff
@@ -41,6 +46,25 @@ function signReqCF(url) {
     return xhrInfo;
 }
 
+function signReqHC(url) {
+    let reqUrl = url.href;
+    const isIssuerUrl = ACTIVE_CONFIG["issue-action"]["urls"]
```
alxdavids: As above, I would personally define `ACTIVE_CONFIG["issue-action"]["urls"]` in background.js and export it here.

src/ext/issuance.js
```diff
@@ -41,6 +46,25 @@ function signReqCF(url) {
...
+    if (!isIssuerUrl) {
+        return null;
+    }
+
+    sentTokens[reqUrl] = true;
+    // Generate tokens and create a JSON request for signing
+    let tokens = GenerateNewTokens(TOKENS_PER_REQUEST);
+    const request = BuildIssueRequest(tokens);
+    // Construct info for xhr signing request
+    let xhrInfo = {newUrl: reqUrl, requestBody: `blinded-tokens=${request}&captcha-bypass=true`, tokens: tokens};
```
alxdavids: Is adding `&captcha-bypass=true` to the request body here intentional? In the CF case we add this to the URL as a query parameter.

durch: Yes, due to the way our routes are set up at the moment appending it to the URL ends up with some messy handling, so we opted for the body instead.

src/ext/background.js
```diff
@@ -126,16 +127,17 @@ function processRedirect(details, oldUrl, newUrl) {
...
+
 function validRedirect(oldUrl, redirectUrl) {
     if (oldUrl.includes("http://")) {
         let urlStr = oldUrl.substring(7);
         let valids = VALID_REDIRECTS;
-        for (let i=0; i<valids.length; i++) {
+        for (let i = 0; i < valids.length; i++) {
             let newUrl = valids[i] + urlStr;
-            if (newUrl == redirectUrl) {
+            if (newUrl === redirectUrl) {
```
henrydcase: is it related to this PR?

durch:
@alxdavids thanks for the quick feedback. All off it makes perfect sense, will follow up shortly, with amendments.

durch:
@alxdavids addressed most of the feedback. Still outstanding are the tests and docs, which we'll start on next. I refactored the `if` clause in `beforeSendHeaders`, and it would be great if you could verify that it does what we need it to.

In addition, I realised that adding new keys to the config objects is quite tedious and error prone, so I added config cloning via `lodash.cloneDeep`. I feel it makes the process of adding new keys, as well as creating new configs more streamlined.

Looking forward to your feedback

src/ext/background.js
```diff
@@ -207,10 +215,44 @@ function processHeaders(details, url) {
  */
 function beforeSendHeaders(request, url) {
     // Cancel if we don't have a token to spend
-    if (!DO_REDEEM || !getSpendFlag(url.host) || checkMaxSpend(url.host) || spentUrl[url.href] || isErrorPage(url.href) || isFaviconUrl(url.href) || REDEEM_METHOD != "reload") {
-        return {cancel: false};
+
+    let reqUrl = url.href;
+    let host = url.host
```
henrydcase: for consistency - let's use semicolons at the end

durch: For sure, that is a typo

src/ext/background.js
```diff
@@ -207,10 +209,45 @@ function processHeaders(details, url) {
...
 function beforeSendHeaders(request, url) {
     // Cancel if we don't have a token to spend
-    if (!DO_REDEEM || !getSpendFlag(url.host) || checkMaxSpend(url.host) || spentUrl[url.href] || isErrorPage(url.href) || isFaviconUrl(url.href) || REDEEM_METHOD != "reload") {
+
+    let reqUrl = url.href;
+
+    // No reload method branch
+    if (DO_REDEEM && REDEEM_METHOD === "no-reload" && !isErrorPage(reqUrl) && !isFaviconUrl(reqUrl)) {
+        // check that we're at an URL that can handle redeems
+        const isRedeemUrl = ACTIVE_CONFIG["spend-action"]["urls"]
```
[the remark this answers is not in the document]
durch: Done

src/ext/background.js
```diff
@@ -207,10 +209,45 @@ function processHeaders(details, url) {
...
+            }
+            setSpendFlag(url.host, null);
+            incrementSpentHost(url.host);
+
+            const http_path = request.method + " " + url.pathname;
+            const redemptionString = BuildRedeemHeader(tokenToSpend, url.hostname, http_path);
+            let headers = request.requestHeaders
+            headers.push({name: ACTIVE_CONFIG["spend-action"]["header-name"], value: redemptionString});
+            headers.push({name: ACTIVE_CONFIG["spend-action"]["header-host-name"], value: url.hostname});
+            headers.push({name: ACTIVE_CONFIG["spend-action"]["header-path-name"], value: http_path});
```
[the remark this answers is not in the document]
durch: Done, done, and done

src/ext/browserUtils.js
```diff
@@ -64,18 +64,21 @@ function attemptRedeem(url, respTabId, target) {
...
             // reset timer for resetting vars
             futureReload[respTabId] = url.href;
             timeSinceLastResp = Date.now();
         }
-    } else {
+    } else if (REDEEM_METHOD === "no-reload") {
+    //    pass
+    }
+    else {
         throw new Error("[privacy-pass]: Incompatible redeem method selected.");
```
henrydcase: This "else if() { // pass } " looks weird. Shoudn't this function be implemented differently.
```
/// ideally this check is in separated function
if !(REDEEM_METHOD === "reload" || REDEEM_METHOD === "no-reload") {
  throw new Error();
}

... function is implemented here...

```

src/ext/background.js
```diff
@@ -207,10 +209,45 @@ function processHeaders(details, url) {
  */
 function beforeSendHeaders(request, url) {
     // Cancel if we don't have a token to spend
-    if (!DO_REDEEM || !getSpendFlag(url.host) || checkMaxSpend(url.host) || spentUrl[url.href] || isErrorPage(url.href) || isFaviconUrl(url.href) || REDEEM_METHOD != "reload") {
+
+    let reqUrl = url.href;
+
+    // No reload method branch
+    if (DO_REDEEM && REDEEM_METHOD === "no-reload" && !isErrorPage(reqUrl) && !isFaviconUrl(reqUrl)) {
```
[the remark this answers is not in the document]
durch: Refactored the entire `if` statement, it could definitely use another set of 👀

src/ext/issuance.js
```diff
@@ -5,11 +5,16 @@
...
 /* exported signReqCF */
+/* exported signReqHC */
 /* exported sendXhrSignReq */
 /* export CACHED_COMMITMENTS_STRING */
 
-function xhrDone(readystate) { return readystate == 4; } // readystate == 4 implies that the response has completed successfully
-function xhrGoodStatus(status) {return status == 200; } // we used to check < 300 but we should be more specific
+function xhrDone(readystate) {
+    return readystate === 4;
+} // readystate == 4 implies that the response has completed successfully
```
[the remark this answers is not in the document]
durch: Moved

src/ext/config.js
```diff
@@ -26,18 +26,23 @@ const exampleConfig = {
...
     "spend-action": {
-        "urls": "<all_urls>", // urls that listeners act on
+        "urls": ["<all_urls>"], // urls that listeners act on
         "redeem-method": "", // what method to use to perform redemption, currently we support "reload" for CF.
         "header-name": "challenge-bypass-token", // name of header for sending redemption token
+        "header-host-name": "challenge-bypass-host", // needed for no-reload method
+        "header-path-name": "challenge-bypass-path", // needed for no-reload method
+    },
+    "issue-action": {
+        "urls": ["<all_urls>"],
```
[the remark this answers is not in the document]
durch: Reworked config, now allows for cloning Config objects, let me know what you think of it.

src/ext/config.js
```diff
@@ -53,57 +63,38 @@ const exampleConfig = {
...
+cfConfig.dev = false;
+cfConfig["max-tokens"] = 300;
+cfConfig["var-reset-ms"] = 2000;
+cfConfig.commitments = "CF"
+cfConfig["spending-restrictions"]["status-code"] = [403,];
+cfConfig["spend-action"]["redeem-method"] = "reload";
+cfConfig["issue-action"]["tokens-per-request"] = 30;
+cfConfig.cookies["clearance-cookie"] = "cf_clearance";
+cfConfig["captcha-domain"] = "captcha.website";
+cfConfig["send-h2c-params"] = true
```
henrydcase: ;

src/ext/background.js
durch: Done, could use some 👀

src/ext/config.js
```diff
@@ -53,57 +63,38 @@ const exampleConfig = {
...
-        "method": "increment",
-    },
-    "send-h2c-params": true
-};
+let cfConfig = lodash.cloneDeep(exampleConfig);
+cfConfig.id = 1;
+cfConfig.dev = false;
+cfConfig["max-tokens"] = 300;
+cfConfig["var-reset-ms"] = 2000;
+cfConfig.commitments = "CF"
```
henrydcase: ;

src/ext/browserUtils.js
durch: Definitely opened to suggestions here, I'd like to avoid throwing an error as the `no-reload` method is valid, there is just no work for it here.

The `// pass` syntax is what allows it to pass `eslint`

src/ext/listeners.js
```diff
@@ -17,7 +17,7 @@ let LISTENER_URLS = ACTIVE_CONFIG["spend-action"]["urls"];
 
 chrome.webRequest.onCompleted.addListener(
     function(details) { handleCompletion(details); },
-    { urls: [LISTENER_URLS] },
+    { urls: ["<all_urls>"] },
```
henrydcase: can you make it const? I mean 

```
const LISTENER_ALL_URLS = "<all_urls>"
```
and then use the const

src/ext/utils.js
```diff
@@ -0,0 +1,45 @@
...
+
+"use strict"
+
+
+/**
+ * Convert URL pattern to regex
+ * @param URL pattern, formatted like those pased to listeners
+ */
+function patternToRegExp(pattern) {
+    if (pattern === "<all_urls>") return /^(?:http|https|file|ftp):\/\/.*/;
```
henrydcase: use ```LISTENER_ALL_URLS ```

src/ext/browserUtils.js
durch: Took a stab at it, generalised an `isValidRedeemMethod`, that checks configs for valid methods, let me know what you think.

henrydcase:
LGTM

src/ext/background.js
```diff
@@ -207,10 +216,44 @@ function processHeaders(details, url) {
...
+                return {requestHeaders: headers};
+            }
+        } else if (REDEEM_METHOD === "reload" && getSpendFlag(host) && !checkMaxSpend(host) && !spentUrl[reqUrl]) {
+            return getReloadHeaders(request, url);
+        }
     }
-    return getReloadHeaders(request, url);
+
+    return {cancel: false}
+
```
alxdavids: Tiny nit: could we remove the empty line here to be consistent with other functions?

src/ext/browserUtils.js
```diff
@@ -185,6 +191,13 @@ function isFaviconUrl(url) {
     return url.includes("favicon");
 }
 
+function isValidRedeemMethod(method) {
```
alxdavids: nice!

durch: thank you :)

durch:
@alxdavids updated CONFIG.md, tests coming next.

alxdavids:
Will review now, looks like you need to run `yarn lint --fix` to fix the CI errors

src/ext/config.js
```diff
@@ -21,7 +21,7 @@ const exampleConfig = {
     "dev": true, // sets whether the configuration should only be used in development
     "sign": true, // sets whether tokens should be sent for signing
     "redeem": true, // sets whether tokens should be sent for redemption
-    "max-spends": 3, // for each host header, sets the max number of tokens that will be spent
+    "max-spends": 3, // for each host header, sets the max number of tokens that will be spent, 0 for unlimited
```
alxdavids: Happy to have this ignored, but perhaps `max-spends` should be left undefined for infinite? Then we can just check `!SPEND_MAX` rather than `SPEND_MAX === 0`. My first thought when I saw `hcConfig["max-spends"] = 0` was that no spends were allowed.

test/beforeHeadersSend.test.js
```diff
@@ -29,172 +32,241 @@ let setMock;
...
+
+            details = {
+                method: "GET",
+                requestHeaders: [],
+                requestId: "212",
+                tabId: "101"
+            };
+            url = new URL(EXAMPLE_HREF);
+            setMockFunctions();
+            setConfig(config.id); // set the CF config
```
alxdavids: remove old comment

test/beforeHeadersSend.test.js
```diff
@@ -29,172 +32,241 @@ let setMock;
...
+                url = new URL(newUrl);
+                setSpendFlag(url.host, true);
+                let redeemHdrs = beforeSendHeaders(details, url);
+                expect(redeemHdrs.cancel).toBeFalsy();
+                expect(redeemHdrs.requestHeaders).toBeFalsy();
+            });
+            test(`max spend has been reached, config.id = ${config.id}`, () => {
+                setSpendFlag(url.host, true);
+                setSpentHosts(url.host, 31);
+                switch (config.id) {
```
alxdavids: Is this switch required? I would have thought that `SPEND_MAX` would be reset when `setConfig` was called in the `beforeEach` block above? So you should be able to just run `workflow.__get__("SPEND_MAX");`

test/beforeHeadersSend.test.js
```diff
@@ -29,172 +32,241 @@ let setMock;
...
+                }
+                expect(reqHeaders).toBeTruthy();
+                let headerName = workflow.__get__("HEADER_NAME");
+                expect(reqHeaders[0].name === headerName).toBeTruthy();
+                expect(reqHeaders[0].value === b64EncodedToken).toBeTruthy();
+            });
+        });
+    })
+
+function setReedemMethod(config_id) {
```
alxdavids: As above, (possibly an artifact of old code) this could probably be handled using `workflow.__get__("REDEEM_METHOD")` because `setConfig` runs before each test.

test/beforeHeadersSend.test.js
```diff
@@ -29,172 +32,241 @@ let setMock;
...
+                setSpentHosts(url.host, 0);
+                setSpentUrl(url.href, false);
+                workflow.__set__("REDEEM_METHOD", "invalid");
+                let redeemHdrs = beforeSendHeaders(details, url);
+                expect(redeemHdrs.cancel).toBeFalsy();
+                expect(redeemHdrs.requestHeaders).toBeFalsy();
+            });
+            test(`no token to spend, config.id = ${config.id}`, () => {
+                localStorage = {
+                    "cf-bypass-tokens": `{}`,
```
alxdavids: I know this is old code, but I guess this should actually be `bypass-tokens-${config.id}` and `bypass-tokens-count-${config.id}`.

test/beforeHeadersSend.test.js
```diff
@@ -29,172 +32,241 @@ let setMock;
...
+                let reqHeaders = redeemHdrs.requestHeaders;
+                expect(getSpendFlag(url.host)).toBeFalsy();
+                expect(getSpendId([details.requestId])).toBeTruthy();
+                expect(getSpentUrl([url.href])).toBeTruthy();
+                switch (config.id) {
+                    case 1:
+                        expect(getSpentTab([details.tabId]) == url.href).toBeTruthy();
+                        break
+                    case 2:
+                        expect(getSpentTab([details.tabId]) == url.href).toBeFalsy();
```
alxdavids: A comment to explain why these are different?

src/ext/config.js
durch: Great idea!

test/beforeHeadersSend.test.js
durch: I'm still trying to grok how exactly `rewire` works, simple `__get__` does not seem to work. That being said, I've started a `lets-get-functional` branch that changes almost  all of the globals to function calls, it feels nicer that way. I also refactored tests there to use `jest-each`, which is much nicer than what I'm doing here. Anyway after we get this merged. I'll submit that PR, I expect to do a lot of work around tests in general there, extract globals and cleanup in general.

test/beforeHeadersSend.test.js
durch: In this setup it does not help, hoping to address that with a general refactor of tests.

test/beforeHeadersSend.test.js
durch: They actually shouldn't be, added tab accounting for hCaptcha as well

src/ext/config.js
```diff
@@ -9,35 +9,44 @@
...
+
 const exampleConfig = {
     "id": 0, // unique integer identifying each individual config
     "dev": true, // sets whether the configuration should only be used in development
     "sign": true, // sets whether tokens should be sent for signing
     "redeem": true, // sets whether tokens should be sent for redemption
-    "sign-reload": true, // whether pages should be reloaded after signing tokens (e.g. to immediately redeem a token)
-    "sign-resp-format": "string", // formatting of response to sign request (string or json)
-    "max-spends": 3, // for each host header, sets the max number of tokens that will be spent
+    "max-spends": 3, // for each host header, sets the max number of tokens that will be spent, 0 for unlimited
```
alxdavids: Nit: can we change this comment to say `undefined` for unlimited.

alxdavids:
Okay everything looks fine to me! Once that last comment is addressed I will merge.

src/ext/config.js
durch: Great catch, fixed!

PR #82, opened 2019-03-14
Automatic generation of commitments

PizzaWhisperer:
This PR addresses the problem of the manual commitment generation. However, due to time constraints, this is still a draft and does not implement every functionality nor is well tested. 

Attempt at a list of that needs to be rethought, enhanced, or even redone:
- Allow concatenation of commitments in the JSON instead of overwriting them
- Generate different commitments for CF and HC, and change the JSON in consequence
- Encode a point such that we have compatibility with existing JSON
- Generate public/private key pair and sign the commitment file
- Upload the JSON somewhere accessible by clients (gRPC?)
- Tests (!)

### DIFF
--- a/beacon/beacon.go
+++ b/beacon/beacon.go
@@ -1,0 +1,137 @@
+package beacon
+
+import (
+	"crypto/elliptic"
+	"crypto/rand"
+	"fmt"
+	"math/big"
+	"sync"
+	"time"
+)
+
+var curve = elliptic.P256()
+var G = BasePoint()
+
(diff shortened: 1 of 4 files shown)


alxdavids:
I would like to keep the commitment generation separate to the extension (this is why we moved the commitments to https://github.com/privacypass/ec-commitments). Having automatic key rotation would be useful, so if you would like to open this PR against the commitment repository (with fixes for the issues that you mentioned) then that would be fine. For now, I'm going to close the PR here.

PR #86, opened 2019-03-21
More restrictive linting

### LINKED ISSUE #85: More specific linting
Enforce style conventions

* indentation
* semi-colons
* function documentation
* trailing spaces

etc...

alxdavids:
Fixes #85 by extending the eslint-config-google linting configuration. Enforces JSDoc, const declarations, semi-colons, etc. We do not enforce JSDoc in the test files.


src/ext/browserUtils.js
```diff
@@ -127,7 +145,11 @@ function isCookieForTab(hrefs, cookieDomain) {
     return false;
 }
 
-// Tor seems to have an object here whereas chrome/firefox just have an id
```
Lekensteyn: Would it make sense to move this comment inside the function such that the justification for this function is maintained?

src/ext/issuance.js
```diff
@@ -39,56 +48,60 @@ function signReqCF(url) {
...
 /**
  * Sends an XHR request containing a BlindTokenRequest for signing a set of tokens
- * @param {object} xhrInfo Object containing information for the XHR that will
- * be returned.
+ * @param {Object} xhrInfo XHR info for asynchronous issuance request
  * @param {URL} url URL object
- * @param {int} tabId Tab ID for the current request
- * @returns {XMLHttpRequest}
+ * @param {Number} tabId Tab ID for the current request
+ * @return {Object}
```
Lekensteyn: XMLHttpRequest seems more accurate than Object, did the linter complain about the former?

test/beforeRequest.test.js
```diff
@@ -125,22 +125,22 @@ function mockXHRCommitments() {
     mockXHR(this);
     this.status = 200;
     this.readyState = 4;
-    this.responseText = `{"CF":{"dev":{"G": "` + testDevG + `","H": "` + testDevH + `"},"1.0":{"G":"` + testG + `","H":"` + testH + `"},"1.1":{"G":"new_11_commitment_g","H":"new_11_commitment_h"}}}`;
+    this.responseText = "{\"CF\":{\"dev\":{\"G\": \"" + testDevG + "\",\"H\": \"" + testDevH + "\"},\"1.0\":{\"G\":\"" + testG + "\",\"H\":\"" + testH + "\"},\"1.1\":{\"G\":\"new_11_commitment_g\",\"H\":\"new_11_commitment_h\"}}}";
```
Lekensteyn: I guess that the linter was complaining about concatenation of template literals. Have you tried something like:
```js
this.responseText = `{"CF":{"dev":{"G":"${testDevG}", ...`
```

I think that would be more readable than escaping quotes. Alternatively, you could use `JSON.stringify` on the object literal.

The same comment applies to storedTokens below (and possibly other places).

alxdavids:
@Lekensteyn updated!

Issue #83, opened 2019-03-19
Support HKDF for DLEQ proof verification


alxdavids:
We currently only support SHAKE but we should also support SHA2-based PRNGs via HKDF (in case that SHAKE is not available at the server-side).

alxdavids:
This is now done.

PR #89, opened 2019-03-26
Adds support for HKDF in DLEQ proof verification

alxdavids:
For #83. We only currently support SHAKE for verifying DLEQ proofs but we add support here for HKDF in case the server does not support SHA3-based primitives. The server specifies the PRNG that the client should use in the IssueResponse (default is SHAKE). The HKDF implementation is taken from https://mozilla.github.io/fxa-js-client/files/client_lib_hkdf.js.html.

Also:
* updated documentation for communicated proof
* fixed indentation for switch cases


src/crypto/local.js
```diff
@@ -373,6 +394,43 @@ function getSeedPRNG(chkM, chkZ, pointG, pointH) {
...
+ * @param {bitArray} salt Salt
+ * @param {sjcl.hash} hash hash function
+ * @return {bitArray}
+ */
+function evaluateHkdf(ikm, length, info, salt, hash) {
+    const mac = new sjcl.misc.hmac(salt, hash);
+    mac.update(ikm);
+    const prk = mac.digest();
+
+    const hashLength = sjcl.bitArray.bitLength(prk)/8;
```
henrydcase: if ``sjcl.bitArray.bitLength(prk)`` can be < 8 then this is a bug

should be "(sjcl.bitArray.bitLength(prk)+7)/8"

alxdavids: fixed

docs/EXT_PROTOCOL_IMPL.md
```diff
@@ -66,15 +65,13 @@ along with a batched DLEQ proof evaluated over all the tokens.
...
     ```
 
-    where: `P` is a `DLEQProof` object; `M`, `Z` are the arrays of points that
-    are received from the client and then after computed after server signing,
-    respectively; `C` is an array of byte arrays that are used in proof
-    verification.
+    where `P` is a `DLEQProof` object.
+
+    Remark: All extra information sent by the server should be ignored. In
+    addition, in a future change we hope to remove the arbirary
```
henrydcase: something missing

alxdavids: fixed

src/crypto/local.js
```diff
@@ -373,6 +394,43 @@ function getSeedPRNG(chkM, chkZ, pointG, pointH) {
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * @param {bitArray} ikm Initial keying material
+ * @param {integer} length Length of the derived key in bytes
+ * @param {bitArray} info Key derivation data
+ * @param {bitArray} salt Salt
+ * @param {sjcl.hash} hash hash function
+ * @return {bitArray}
+ */
+function evaluateHkdf(ikm, length, info, salt, hash) {
```
henrydcase: would be good to have a test with few KAT's for this

src/crypto/local.js
```diff
@@ -373,6 +394,43 @@ function getSeedPRNG(chkM, chkZ, pointG, pointH) {
     return sjcl.codec.hex.fromBits(h.finalize());
 }
 
+/**
+ * hkdf - The HMAC-based Key Derivation Function
+ * based on https://github.com/mozilla/node-hkdf
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
henrydcase: I think it would be good to mention that this doesn't handle a case when ``len(salt)==0``
From https://tools.ietf.org/html/rfc5869
```

      salt     optional salt value (a non-secret random value);
               if not provided, it is set to a string of HashLen zeros.
```

src/crypto/local.js
```diff
@@ -373,6 +394,43 @@ function getSeedPRNG(chkM, chkZ, pointG, pointH) {
...
+
+    const hashLength = (sjcl.bitArray.bitLength(prk)+7)/8;
+    const numBlocks = Math.ceil(length / hashLength);
+    let prev = sjcl.codec.hex.toBits("");
+    let output = "";
+    for (let i = 0; i < numBlocks; i++) {
+        const hmac = new sjcl.misc.hmac(prk, hash);
+        const input = sjcl.bitArray.concat(
+            sjcl.bitArray.concat(prev, info),
+            sjcl.codec.utf8String.toBits((String.fromCharCode(i + 1)))
```
henrydcase: just to make sure - this will always be encoded on 8 bits, right?

alxdavids: There was an exception if it was over 8 bits, but I've added a more explicit error

alxdavids:
Thanks! I've added KATs for HKDF

src/crypto/local.js
```diff
@@ -307,51 +308,71 @@ function verifyProof(proofObj, tokens, signatures, commitments) {
...
+        case "hkdf":
+            out = sjcl.codec.hex.fromBits(prng.func(sjcl.codec.hex.toBits(seed), (bitLen+7)/8, sjcl.codec.utf8String.toBits("DLEQ_PROOF"), salt, CURVE_H2C_HASH));
+            break;
+        default:
+            throw new Error(`Server specified PRNG is not compatible: ${prng.name}`);
     }
-    return rnd;
+    // Masking is not strictly necessary for p256 but better to be completely
+    // compatible in case that the curve changes
+    const h = "0x" + out.substr(0, 2);
```
Lekensteyn: Oh, this looked strange at first, but it seems that you want to convert a hex value to a number. That could also be done with `parseInt(out.substr(0, 2), 16)` which would return a Number instead of String.

(The code existed before, it is not really wrong, but mixing types seemed odd.)

test/crypto.test.js
```diff
@@ -0,0 +1,60 @@
...
+    l: 42,
+    output: `0x8da4e775a563c18f715f802a063c5a31b8a11f5c5ee1879ec3454e5f3c738d2d9d201395faa4b61a96c8`,
+}];
+
+describe("HKDF", () => {
+    describe("KATs", () => {
+        testVectors.forEach((vector) => {
+            test(`${vector.name}`, () => {
+                const out = evaluateHkdf(vector.ikm, vector.l, vector.info, vector.salt, vector.hash);
+                expect(checkHexOutput(sjcl.codec.hex.fromBits(out), vector.output)).toBeTruthy();
```
Lekensteyn: Drop the template string?

    `${vector.name}`
->

    vector.name.

Maybe use something like this:

    expect(sjcl.codec.hex.fromBits(out)).toEqual(vector.output);

and drop the 0x from the testVectors above? (Also, consider using normal strings there)

Lekensteyn:
Also, I did notice that "blind" is a hex number written as string whereas other values use base64. Would it make sense to use base64 for blind as well or would it somehow break backwards compat?
```js
    const storedTokens = `[ { "data":[24,62,56,102,76,127,201,111,161,218,249,109,34,122,160,219,93,186,246,12,178,249,241,108,69,181,77,140,158,13,216,184],"point":"/MWxehOPdGROly7JRQxXp4G8WRzMHTqIjtc17kXrk6W4i2nIp3QRv3/1EVQAeJfmTvIwVUgJTMI3KhGQ4pSNTQ==","blind":"0x46af9794d53f040607a35ad297f92aef6a9879686279a12a0a478b2e0bde9089"},{"token":[131,120,153,53,158,58,11,155,160,109,247,176,176,153,14,161,150,120,43,180,188,37,35,75,52,219,177,16,24,101,241,159],"point":"sn4KWtjU+RL7aE53zp4wUdhok4UU9iZTAwQVVAmBoGA+XltG/E3V5xIKZ1fxDs0qhbFG1ujXajYUt831rQcCug==","blind":"0xd475b86c84c94586503f035911388dd702f056472a755e964cbbb3b58c76bd53" } ]`;
```

alxdavids:
> Also, I did notice that "blind" is a hex number written as string whereas other values use base64. Would it make sense to use base64 for blind as well or would it somehow break backwards compat?
> 
> ```js
>     const storedTokens = `[ { "data":[24,62,56,102,76,127,201,111,161,218,249,109,34,122,160,219,93,186,246,12,178,249,241,108,69,181,77,140,158,13,216,184],"point":"/MWxehOPdGROly7JRQxXp4G8WRzMHTqIjtc17kXrk6W4i2nIp3QRv3/1EVQAeJfmTvIwVUgJTMI3KhGQ4pSNTQ==","blind":"0x46af9794d53f040607a35ad297f92aef6a9879686279a12a0a478b2e0bde9089"},{"token":[131,120,153,53,158,58,11,155,160,109,247,176,176,153,14,161,150,120,43,180,188,37,35,75,52,219,177,16,24,101,241,159],"point":"sn4KWtjU+RL7aE53zp4wUdhok4UU9iZTAwQVVAmBoGA+XltG/E3V5xIKZ1fxDs0qhbFG1ujXajYUt831rQcCug==","blind":"0xd475b86c84c94586503f035911388dd702f056472a755e964cbbb3b58c76bd53" } ]`;
> ```

Good point, I think we can change this to base64 without breaking backwards compatibility. I'll make a separate issue for it.

src/crypto/local.js
alxdavids: This is handled, see KATs

Issue #91, opened 2019-03-28
Provide support for compressed points in issuance responses


alxdavids:
* Accept signed points that are in compressed encoding when returned by server.
* Validate DLEQ proof using compressed bytes rather than uncompressed

PR #92, opened 2019-03-29
Add support for compressed elliptic curve points in issuance responses

alxdavids:
For #91:
* Points can be received in compressed encoding from the server
* DLEQ proofs are verified using the compression used by the server


Lekensteyn:
Point compression is reasonable to reduce the size for transport and storage. I have some questions though:
- Is there a reason to support PointData with both uncompressed and compressed points? These are only used internally during computations, if only uncompressed points are supported (as was the case before), then I think the code could be simplified (no more need to pass the "compressed" flag all over the place.
- Is there a reason to forbid mixing compressed/uncompressed points?
- The uncompressed points are already encoded with a prefix according to Section 2.3.3 of [SEC1](https://www.secg.org/sec1-v2.pdf). So it is unambiguous whether a point is compressed or not. What about extending sec1DecodePoint and sec1EncodePoint with support for the compressed forms (prefix 3 and 4) instead of making its callers responsible for deciding whether to call the uncompressed/compressed point decoding/encoding function?

alxdavids:
Thanks for the feedback @Lekensteyn, my thoughts:
- It's true that we could support internal operations using uncompressed points only, but I think it is easier to comprehend if we provide data to functions in the closest form to that it is received in. The internal computations are mirrored server-side, and so potential new server integrations don't have to switch points to uncompressed encodings before computing stuff (if they only want to support compressed points). Having said that, I agree that passing the `compressed` flag around is annoying.
- There is no real reason to forbid mixing encodings, but I think this would be harder to support for little gain?
- I think this is a good idea, I'll update these functions to do that. It may also help to reduce the number of times the `compressed` flag is required.

Lekensteyn:
Going forward I guess that compressed points in PointData would be preferred over uncompressed points due to the space savings? Support for uncompressed points would then exist for backwards compatibility reasons only. (And indeed, `compressed` cannot easily be removed... :/).

For simplicity sake, would it be possible and beneficial to support only compressed points for hkdf (which does not needs backwards compatibility since it is new)?

armfazh:
> For simplicity sake, would it be possible and beneficial to support only compressed points for hkdf (which does not needs backwards compatibility since it is new)?

This is a good idea for reducing bandwidth.But it adds the cost for decompressing points. There is trade-off between time and space. Hence, it think the most flexible option is to allow both compressed and decompressed points.

alxdavids:
@Lekensteyn Updated the existing encoding/decoding functions to handle compressed points as well instead of handling it in the callers. 
Following @armfazh comments, I've kept stored points uncompressed to avoid computational overheads. This can be changed in the future without affecting backwards-compatibility if it turns out that we can afford it.

src/crypto/local.js
```diff
@@ -141,8 +198,11 @@ function compressPoint(P) {
  */
 function decompressPoint(bytes) {
     const yTag = bytes[0];
+    const expLength = Math.ceil(sjcl.bitArray.bitLength(CURVE.r.toBits())/8)+1;
```
henrydcase: if ``sjcl.bitArray.bitLength(CURVE.r.toBits()) == k*8`` then this will result in (k+1) bytes. Seems to me check at line 204 will fail

henrydcase: ah,, no - ignore. that's OK

src/crypto/local.js
```diff
@@ -270,27 +313,27 @@ function verifyProof(proofObj, tokens, signatures, commitments, prngName) {
...
+    const pointG = sec1DecodeFromBase64(commitments.G);
+    const pointH = sec1DecodeFromBase64(commitments.H);
 
     // Recompute A and B for proof verification
     const cH = _scalarMult(dleq.C, pointH);
     const rG = _scalarMult(dleq.R, pointG);
     const A = cH.toJac().add(rG).toAffine();
 
-    const composites = recomputeComposites(chkM, chkZ, pointG, pointH, prngName);
+    const composites = recomputeComposites(chkM, chkZ, pointG, pointH, compressed, prngName);
```
henrydcase: nit: Passing this ``compressed`` flag doesn't look perfect. If the point is compressed or not is a property of the point  and hence, implementation of a point should have such property. This way you can avoid changing at least ``recomputeComposites`` function (in general - any function which will be only passing a ``Point``)

I would introduce structure wrapping sjcl.ecc.point and add "compressed" field there.

alxdavids: I agree the `compressed` flag is annoying. However, I'm slightly against defining an extra struct here because I'm not sure it will help to make things less ambiguous and I would like to enforce that points are received in exactly the same format for an entire server response. So passing around a compression flag for each point seems like overkill.

The latest change removes the flag from most of the function calls, but the compressed flag is still defined with respect to the entire set of tokens received in the response.

src/crypto/local.js
henrydcase: Is there a reason for not simply using CURVE.r.bitLength()/8 + 1?

alxdavids: Good point, we can just bitlength on the sjcl.bn type, I hadn't noticed that. In terms of the `+1` vs `Math.ceil`, I think Math.ceil is a bit more idiomatic and makes it clear that the result of the division may be a non-integer

alxdavids: Actually Math.ceil is required because we need to get the number of bytes + 1

henrydcase: I don't think so. From what I can see, is that bitLength() returns number of bits, but roudned up to nearest byte. So Math.ceil is useless.
You need +1 because expectedLength is a length of prime order + 1 byte for "compression flag"

alxdavids: Ah yep, you're correct, it is useless!

PR #94, opened 2019-04-05
Fix some issues with Chrome & sub-resources

alxdavids:
Using Privacy Pass in Chrome sometimes failed to load certain resources because the responseHeaders returned in the webRequest API were empty. This meant that no token was ever spent for them. This was a problem for some CF websites, particularly if sub-resources were stored off the initial requested domain.

This change provides a fix by sending an async request to a CF endpoint for any resource where this occurs and sets a spend flag if the website returns with the correct header and status code for the active configuration. I've added some config options for doing this that are off by default (but on for the CF config).

Other changes:
* fixed a config mix-up for naming verification & connection errors
* only reload tabs with id >= 0
* remove SPEND_IFRAME tests because that config option is not used anymore

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -131,7 +131,20 @@ The name of the header that contains a hostname that is sending the redemption r
 
 #### config["spend-action"]["header-path-name"]
 
-The name of the header that contains a path thatis sending the redemption request.
+The name of the header that contains a path that is sending the redemption
+request.
+
+#### config["spend-action"]["empty-resp-headers"]
+
+If an empty set of response headers is received with the correct status code (as
+defined in `config["spending-restrictions"]["status-code"]`) then this object
+contains an array of strings that correspond to possible ways of acquiring the
+headers. Currently we only support `"direct-request"`, which sends a direct
+request to the same URL with a specific endpoint attached as defined in
+`config["opt-endpoints"]["challenge"]`.
+
+This option was introduced to mitigate problems with Chrome in conjunction with
+sub-resources hosted on separate Cloudflare domains.
 
 ### config["issue-action"]
 
@@ -183,6 +196,10 @@ by the provider. In the case of Cloudflare, this is `"cf_clearance"`.
 A string specifying a domain where users can obtain signed tokens by solving a
 challenge/CAPTCHA. This is helpful to allow users to build up initial stockpiles
 of tokens before they browse.
+
+### config["opt-endpoints"]
+
+Optional endpoints for use by the particular configuration.
 
 ### config["error-codes"]
 
--- a/jest/globals.js
+++ b/jest/globals.js
@@ -55,7 +55,7 @@ window.workflowSet = () => {
     workflow.__set__("setSpendFlag", setSpendFlagMock);
     workflow.__set__("getSpendFlag", getSpendFlagMock);
 
-    workflow.__set__("spentUrlMock", spentUrlMock);
+    workflow.__set__("spentUrl", spentUrlMock);
     workflow.__set__("setSpentUrl", setSpentUrlMock);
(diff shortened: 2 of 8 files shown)


src/ext/background.js
```diff
@@ -195,9 +199,62 @@ function processHeaders(details, url) {
...
+    if (details.responseHeaders.length === 0 && SPEND_STATUS_CODE.includes(details.statusCode)) {
+        const xhr = new XMLHttpRequest();
+        xhr.onreadystatechange = function() {
+            // We return a boolean for testing purposes
+            let xhrRet = false;
+            if (this.readyState === this.HEADERS_RECEIVED) {
+                if (SPEND_STATUS_CODE.includes(xhr.status) && xhr.getResponseHeader(CHL_BYPASS_SUPPORT) === CONFIG_ID) {
+                    // don't return anything here because it is async
+                    decideRedeem(details, url);
+                    xhr.abort();
```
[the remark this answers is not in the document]
alxdavids: Nope, I'll remove

src/ext/background.js
```diff
@@ -195,9 +199,62 @@ function processHeaders(details, url) {
         }
```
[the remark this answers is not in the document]
alxdavids: True, I can remove `activated` and just move the if below into this loop and break afterwards.

src/ext/background.js
```diff
@@ -191,13 +194,59 @@ function processHeaders(details, url) {
...
+    }
+
+    return ret;
+}
+
+/**
+ * Try a direct request against a challenge endpoint if the response headers are
+ * empty. This fixes some strange behaviour with CF sites and Chrome.
+ * @param {Object} details
+ * @param {URL} url
```
henrydcase: no description for params

jest/globals.js
```diff
@@ -121,14 +121,15 @@ window.updateBrowserTabMock = jest.fn();
...
 
+window.getMock = (key) => JSON.parse(localStorage.getItem(key));
+window.setMock = (key, value) => localStorage.setItem(key, JSON.stringify(value));
+
 window.CACHED_COMMITMENTS_STRING = "cached-commitments";
 window.clearCachedCommitmentsMock = () => setMock(CACHED_COMMITMENTS_STRING, null);
 
-window.setSpendFlagMock = (key, value) => setMock(key, value);
-window.getSpendFlagMock = (key) => getMock(key)
+window.setSpendFlagMock = (key, value) => setMock(key, value)
```
armfazh: semi-colons at the end.

src/ext/config.js
```diff
@@ -88,6 +90,9 @@ function PPConfigs() {
     cfConfig.cookies["clearance-cookie"] = "cf_clearance";
     cfConfig["captcha-domain"] = "captcha.website";
     cfConfig["send-h2c-params"] = true;
+    cfConfig["opt-endpoints"].challenge = "/cdn-cgi/challenge";
+    cfConfig["spend-action"]["empty-resp-headers"] = ["direct-request"];
+    cfConfig["max-spends"] = 1; // this used to be 3 but lead to more spends than necessary so I've reduced it to 1 and it doesn't seem to have an impact
```
armfazh: In the example config should be 1 too?
What are the pros/cons of setting this value to 1 or higher?

alxdavids: Yes it probably should, good point. The "max-spends" rule just limits the number of tokens that are spent for a single host at any time, it is reset after a spend is completed.

docs/CONFIG.md
```diff
@@ -131,7 +131,20 @@ The name of the header that contains a hostname that is sending the redemption r
 
 #### config["spend-action"]["header-path-name"]
```
thaimacao887: #97

Issue #96, opened 2019-04-25
Cant' tell if its working or not ...


ghost:
Seeing as I'm using a heavily modded user.js and many privacy extensions, I'm not sure PrivacyPass has the right permissions or APIs enabled to properly work ... I get 30 passes when solving captcha.website but still getting 5+ reCAPTCHA loops on other sites.

alxdavids:
Are these reCAPTCHAs that are being shown on Cloudflare-protected websites? Privacy Pass doesn't prevent seeing reCAPTCHAs in general, only those that are hosted on the Cloudflare challenge page.

sesam:
I finally get to to CAPTCHAs on https://captcha.website/ and I did get 30 passes - yay!
But reloading the page, even after trying "Hard reload and remove cache" (a somewhat hidden developer feature in Chrome) I still get the same captcha page, and can't seem to get past it to see the page that used to be there when captchas were not being shown to me. Hm.

alxdavids:
Privacy Pass deliberately avoids spending tokens on captcha.website to prevent token wastage. The extension is currently only intended to improve the browsing experience for users who experience Cloudflare CAPTCHAs (until we support more use-cases).

I'm going to close this issue as the OP hasn't updated with more information. We can reopen if they do.

PR #100, opened 2019-05-07
Update README.md

sesam:
- looked in .circleci and then clarified what "(ci) PASSED" means in this particular case
- added text meant to be seen "above the fold" when a first-time visitor looks at the github page
- clarified "protocol we use", hopefully correctly?

(Also BTW, I get captchas now on captcha.website, so I'll get the extension activated again and see if this works for me too now! I'm eager to test it!)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Challenge Bypass Extension
 
-[![CircleCI](https://circleci.com/gh/privacypass/challenge-bypass-extension.svg?style=svg)](https://circleci.com/gh/privacypass/challenge-bypass-extension)
+Tests: [![CircleCI](https://circleci.com/gh/privacypass/challenge-bypass-extension.svg?style=svg)](https://circleci.com/gh/privacypass/challenge-bypass-extension)
 
 The Privacy Pass extension implements the Privacy Pass protocol for providing a
 private authentication mechanism during web browsing. Privacy Pass is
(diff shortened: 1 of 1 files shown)


sesam:
@armfazh Please, can you take a minute and say if this PR needs modification?

Maybe the text "The extension is currently only intended to improve the browsing experience for users who experience Cloudflare CAPTCHAs (until we support more use-cases)." (as seen in #96) could also be good to add, though I'm not sure where it could be suitable.

README.md
```diff
@@ -1,6 +1,6 @@
 # Challenge Bypass Extension
 
-[![CircleCI](https://circleci.com/gh/privacypass/challenge-bypass-extension.svg?style=svg)](https://circleci.com/gh/privacypass/challenge-bypass-extension)
+Tests: [![CircleCI](https://circleci.com/gh/privacypass/challenge-bypass-extension.svg?style=svg)](https://circleci.com/gh/privacypass/challenge-bypass-extension)
```
alxdavids: I don't think we need to clarify this.

README.md
```diff
@@ -11,7 +11,15 @@ and [Firefox](https://addons.mozilla.org/en-US/firefox/addon/privacy-pass/)
 (v48+). An example server implementation that is compatible with this extension
 is available [here](https://github.com/privacypass/challenge-bypass-server).
 
-The protocol we use is based on a realization of a 'Verifiable, Oblivious
+In effect, the extension enables users to less frequently redo CAPTCHAs _from sites
```
alxdavids: This paragraph does not belong up here. If it is necessary to add then it should be in the FAQ since this has only arisen from questions that have been asked in the issues.

README.md
```diff
@@ -11,7 +11,15 @@ and [Firefox](https://addons.mozilla.org/en-US/firefox/addon/privacy-pass/)
...
+implementing the Privacy Pass protocol_. The extension doesn't show if the protocol is
+active or not. Not all users or websites use this, also since this technology is not
+yet well known. Best chance to try this out is if you're in a region that are deemed to
+need CAPTCHAs, and visit a CloudFlare site, like the example http://captcha.website/
+If, by visiting, you don't get any CAPTCHA, you're welcome to hang around anyway and
+learn more about this technology!
+
+The Privacy Protocol is based on a realization of a 'Verifiable, Oblivious
```
alxdavids: Again I'm not sure if this change needs making, but if it does then it should be the `Privacy Pass protocol`.

README.md
alxdavids: Also:

> The extension doesn't show if the protocol is
active or not. Not all users or websites use this, also since this technology is not
yet well known. Best chance to try this out is if you're in a region that are deemed to
need CAPTCHAs, and visit a CloudFlare site, like the example http://captcha.website/
If, by visiting, you don't get any CAPTCHA, you're welcome to hang around anyway and
learn more about this technology!

appear out of place in this repository. I can see the value in adding a clarifier to the FAQ that there are some issues with https://captcha.website that prevent it from working properly, but the rest seems unnecessary.

sesam:
@alxdavids Thank for your time. Are none of the suggested changes welcome? 
Are you interested in any other kind of PR that makes this extension more understandable for users?

I'm still feeling unhappy about the general user experience. I can also respect if this is primarily a tech preview rather than a product meant for beginner users.

alxdavids:
I'm going to close this PR due to inactivity. 

In general, I think we want to keep the README in this repository aimed towards developers and I think this change focuses too much on the extension experience. However, in the near future I may update the extension description in the Chrome and FF stores to reflect some of the FAQs that we get.

PR #110, opened 2019-09-25
No longer minify source code

alxdavids:
Minified source code is allowed by both Chrome and Firefox but we've had some probelms in the past with store guidelines so it is probably just best to present the source code as is.

Changes:
- Switch to make for building & testing
- Had to change a load of variables from `const` to `let` to allow mocking them during tests


alxdavids:
Took a while to fix the CI but it's ready to review now

alxdavids:
@Lekensteyn @armfazh for review

PR #111, opened 2019-09-25
Adds support for verifying signature of commitments

### LINKED ISSUE #114: verify expiration date of commitments.
commitments now include a field that indicates an expiration date.
The extension should test expiration day.

armfazh:
Changes:
 - Configuration file adds a field `pkey-commitments` for storing the
 public key for verifying commitments.
 - If the commitment has a signature field, then proceeds with
 verificiation.
 - Assumes public key is ECDSA P-256 stored in PEM format.
 - Verify commitments expiration date. (Closes #114 )
 - Backwards compatible with version 1.0.

Mentioned in https://github.com/privacypass/ec-commitments/issues/4


src/crypto/local.js
```diff
@@ -17,8 +17,14 @@
 /* exported getCurvePoints */
 /* exported getBigNumFromBytes */
 /* exported getActiveECSettings */
+/* exported verifyCommitments */
+
 "use strict";
 
+const parseKeys = require("parse-asn1");
```
armfazh: I don't know where is the right place to include these packages.

alxdavids: Good point, it might be a good time to start a separate file for requirements that is then included in the building process?

src/ext/issuance.js
```diff
@@ -312,23 +312,18 @@ function verifyProofAndStoreTokens(url, tabId, tokens, issueResp, commitments) {
...
-            commG = comms[version]["G"];
-            commH = comms[version]["H"];
-        }
+    const cmt = comms[version];
+    if (typeof cmt === "undefined") {
+        throw new Error("[privacy-pass]: Retrieved version: " + version + " not available.");
     }
-
-    return {G: commG, H: commH};
+    if (cmt.sig) {
```
alxdavids: I'm a little worried about only verifying if there is a signature field but I'm not sure what the best alternative would be.

Perhaps we could an extra member to each of the configs in config.js that specifies whether signatures should be verified by default, we could set the CF config to default to true, and the HC config to default to false until they provide signatures?

I'm not sure what the best approach to still allowing version `"1.0"` to work for CF would be though. I guess we could provide an extra member in the configs that specifies trusted versions (that would include "1.0" for CF?). This would mean that we would structure the commitments members as:
```
cfConfig["commitments"] = {
  "verification-key": <public-key>,
  "verify": true,
  "trusted": ["1.0"],
}
```
and for hCaptcha:
```
hcConfig["commitments"] = {
  "verification-key": null,
  "verify": false,
}
```
Or we could just provide a signature for version 1.0 in a future change, but that would mean the master branch would not work for a while. Do you have an opinion?

src/crypto/local.js
```diff
@@ -287,6 +293,74 @@ function validResponseCompression(compression, setting) {
...
+    }
+}
+
+/**
+ * Verify the signature of commitments.
+ * @param {json} comms - commitments to verify
+ * @param {string} derPublicKey - A public key in DER format.
+ * @return {boolean} True, if the commitment signature is valid.
+ */
+function verifyCommitments(comms, derPublicKey) {
```
alxdavids: I think we should also verify the expiry date of the commitment in here. I couldn't seem to find it?

test/issuance.test.js
```diff
@@ -84,6 +84,13 @@ describe("commitments parsing and caching", () => {
         expect(v11H === commitments.H).toBeTruthy();
     });
 
+    test("parse correctly (v1.01)", () => {
```
alxdavids: It would be worth adding some tests that should fail, such as: 
- if the signature is incorrect and fails to verify
- if the verification key is invalid so that it fails to parse

src/crypto/local.js
armfazh: I would handle this verification in another PR.

src/crypto/local.js
armfazh: Added a requires.js file.

src/ext/issuance.js
armfazh: I think the configuration is getting a lot of flags, which adds more complexity to the program.
 I would prefer something simpler. I think that the presence of signatures in commitments must be enforced. 
This can be controlled by the version of our implementations.
example: version 1.xx means signature is optional.
 version 2.xx means signature is required.


Assume that v2.0 is a big step from v1.0 then it is ok to work on an alternate branch (for not breaking master). and the release day make the merge with master.

armfazh: You are right. Let's agree on a path to follow.

test/issuance.test.js
```diff
@@ -124,6 +147,14 @@ describe("commitments parsing and caching", () => {
         clearCachedCommitmentsMock();
         expect(getCachedCommitments).not.toThrowError();
     });
+
+    test("bad commitments signature", () => {
+        expect(()=>{
+            const xhr = createVerificationXHR(); // this usually takes params
+            const commitments = retrieveCommitments(xhr, "sig-bad");
+            expect(commitments).toBeUndefined();
```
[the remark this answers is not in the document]
armfazh: True, it was changed.

test/issuance.test.js
```diff
@@ -124,6 +147,14 @@ describe("commitments parsing and caching", () => {
...
+        }).toThrowError("sig");
```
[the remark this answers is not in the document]
armfazh: Done

test/issuance.test.js
armfazh: done

test/issuance.test.js
```diff
@@ -149,12 +149,26 @@ describe("commitments parsing and caching", () => {
         expect(getCachedCommitments).not.toThrowError();
     });
 
-    test("bad commitments signature", () => {
+    test("malformed commitment's signature", () => {
```
alxdavids: nit: commitment's -> commitments

alxdavids:
Okay this change LGTM so far. 

For the last change regarding enforcing of checking signatures, I think the best thing would be to add a new member (`enforce-verify`) to the configs that specifies whether signatures need to be enforced. The signature verification check would then check this value before it proceeded. This setting would default to `false`, but before the release we could switch the Cloudflare value to `true` and this wouldn't affect the other configuration. The member could be added to the the existing `issuance` members as it used only in the issuance phase of the protocol.

What do you think?

armfazh:
For version 2, we enforce signature of commitments. These changes deviate from master, and for this reason this is now a PR for `UpcomingVersionTwo` branch.

alxdavids:
@armfazh I've just realised that using `require` is not allowed in the browser environment, this is why the keccak functionality is added in directly as source code. I suspect that we will have to remove `requires.js` from the browser environment (although we can use for tests, if necessary).

For `Buffer`, I assume we can just re-use the functionality provided by `ArrayBuffer` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays). For `assert`, I believe we can just rewrite the functionality as conditionals. For `parse-asn1` this will be more complicated, but I think we may just have to include the source directly. It looks like the module you're using is this https://git.coolaj86.com/coolaj86/asn1-parser.js/src/branch/master/asn1-parser.js which is 161 lines of code, so it should be fine to just include it directly in the building process and then call the required functions directly.

src/ext/issuance.js
```diff
@@ -312,23 +312,21 @@ function verifyProofAndStoreTokens(url, tabId, tokens, issueResp, commitments) {
...
-            commG = comms["dev"]["G"];
-            commH = comms["dev"]["H"];
-        } else {
-            commG = comms[version]["G"];
-            commH = comms[version]["H"];
+    const cmt = comms[version];
+    if (typeof cmt === "undefined") {
+        throw new Error("[privacy-pass]: Retrieved version: " + version + " not available.");
+    }
+    if (version[0] === "2") {
```
alxdavids: Happy to leave this in for now, but I think we should remove this version check before we merge this branch into master in the future.

alxdavids: Ignore this comment, I think we can probably leave this in the initial release and then officially remove support for version `1` keys in a separate release.

alxdavids:
@armfazh fixed CI with new test data

alxdavids:
This looks good to me now. I've tried it in the browser and everything loads fine. Mergin.

PR #112, opened 2019-09-26
Handle more error codes

alxdavids:
We plan to use new error codes in future server implementations that will highlight specific cases where clients have sent bad requests, or whether an unknown error has occurred. These error codes primarily provide information the client during a failed redemption request.

(PR on top of unminified source change to simplify merging later)


alxdavids:
@Lekensteyn @armfazh for review

PR #113, opened 2019-09-28
Move building to Makefile

armfazh:
This is PR on top of #110. If approved then closes ##110.

I noticed a lot of inconsistencies produced by moving compilation from yarn to Makefile. Because of that, I took some time to clean up these inconsistencies. Now targets of yarn are equivalent to targets using Makefile.

### Summary of Changes:

by #110:
 - No more closure compression.
 - Switch to make for building & testing
 - Had to change a load of variables from const to let to allow mocking them during tests

by this PR:
 - Building logic moved to Makefile
 - package.json maintains only jest and eslint commands.
 - CircleCI split on named phases.
 - Updated readme compilation steps.
 - Linting Markdown files.
 - Removed (apparently false-)dependency on JDK for sjcl.


armfazh:
This PR is ready for review.

PR #115, opened 2019-10-16
Documentation and version updates

alxdavids:
- Documentation
- Menifest version -> 2.0


docs/HASH_TO_CURVE.md
```diff
@@ -24,32 +24,11 @@ for more details.
```
armfazh: Broken link at line 22.

docs/HASH_TO_CURVE.md
```diff
@@ -24,32 +24,11 @@ for more details.
 
 ## Affine SSWU method
 
-In Privacy Pass v1.1 we will move to supporting both the previous
+In the new release of Privacy Pass we will move to supporting both the previous
 `hash-and-increment` method, and a new method that we call "affine SSWU". The
 affine SSWU method always returns a elliptic curve point and thus has zero chance
```
armfazh:
```suggestion
 affine SSWU method always returns an elliptic curve point and thus has zero chance
```

docs/HASH_TO_CURVE.md
```diff
@@ -24,32 +24,11 @@ for more details.
...
 of failure. The method used is the simplified version of SWU, which was proposed by Brier et al. based on the previous work of Shallue, Woestijne and Ulas (hence SSWU).
```
armfazh:
```suggestion
 of failure. The encoding method, proposed by [Brier et al.](http://eprint.iacr.org/2009/340), is a simplified version of SWU encoding, which is based on the previous works of Shallue and Woestijne, and Ulas (hence SSWU).
```

docs/HASH_TO_CURVE.md
```diff
@@ -24,32 +24,11 @@ for more details.
...
- 8. g  = g+B
- 9. y  = g^((p+1)/4)
-10. t0 = y^2
-11. IF t0 ≠ g THEN
-12.   x  = u*x
-13.   y  = (−1)^((p+1)/4)*u*t*y
-14. ENDIF
-15. RETURN (x,y)
-```
+The algorithm is described in <https://datatracker.ietf.org/doc/draft-irtf-cfrg-hash-to-curve/>. The input is a field element `t` that is obtained by hashing the input bytes into GF(p) using the output of a random oracle (we use the hash function SHA-256). The output is a coordinate pair (x,y) for a curve point on the  elliptic curve E(GF(p)).
```
armfazh:
```suggestion
The algorithm is described in <https://datatracker.ietf.org/doc/draft-irtf-cfrg-hash-to-curve/>. The input is a field element `t` that is obtained by hashing the input bytes into GF(p) using the output of a random oracle (we use the hash function SHA-256). The output is a coordinate pair (x,y) for a point on the elliptic curve P-256.
```

PR #118, opened 2019-10-17
Add HC `pkey-commitments`

durch:


alxdavids:
@durch feel free to add unit tests in a separate PR for making sure you can verify your commitments, if you would like.

PR #121, opened 2019-10-18
Fix redirect issue

alxdavids:
Redemptions weren't working for sites that require redirects. This change ups the max spends for CF to allow for spending redemption tokens on redirected domains that don't supply a single-domain cookie until after the spend occurs.


armfazh:
IIUC each redirection accounts for one extra spend?

alxdavids:
@armfazh Yes that is the case, the problem is that we don't get a clearance cookie if a redirect occurs, so we have to spend a token after the redirect has happened.

PR #126, opened 2019-10-22
Fixes an issue when reloading tabs in Chrome for redirects

alxdavids:
- In Chrome, reloading a tab after a redirection would just reload the old URL
- This change preserves redirections and updates the tab instead if a spend has occurred. This uses the new URL for the reload.


armfazh:
I can't reproduce it. :(
Go forward.

PR #132, opened 2019-10-24
HC patch to preserve backwards compatibility

durch:
This patch preserve hcaptca.com compatibility with the `UpcomingVersionTwo`. In addition to that it fixes some problems in token issuance.

Compatibility is preserved by not verifying commitments if the `sig` key is not present in the pulled commitment. Even tough we have generated keys with the new format, the reference implementation of the bypass server does still not support it. So this change should be inline with other compatibility preserving changes made recently.

@alxdavids we'll be ready to switch to the new format as soon as the new bypass-server becomes available, as usual all feedback regarding this PR is welcome. There might be more changes incoming once we verify the redeem path as well.

### DIFF
--- a/src/ext/background.js
+++ b/src/ext/background.js
@@ -446,7 +446,7 @@ function beforeRequest(details, url) {
             xhrInfo = signReqCF(url);
             break;
         case 2:
-            xhrInfo = signReqHC(url);
+            xhrInfo = signReqHC(url, details);
             break;
         default:
             throw new Error("Incorrect config ID specified");
@@ -460,6 +460,9 @@ function beforeRequest(details, url) {
 
     // actually send the token signing request via xhr and return the xhr object
     const xhr = sendXhrSignReq(xhrInfo, url, details.tabId);
+    if (xhrInfo.cancel) {
+        return false;
+    }
     return {xhr: xhr};
 }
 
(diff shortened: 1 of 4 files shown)


alxdavids:
Hi @durch do you mind targeting this PR at master as we're back to working off the main branch now? Also we're not planning on making any changes to the server so that shouldn't be an issue, I believe it should still be compatible with the version of the server that is provided.

I have actually added a new member to config.js that allows you to specify 1.0 commitments which directly in the source code, for those that do not have a signature. See https://github.com/privacypass/challenge-bypass-extension/blob/master/src/ext/config.js#L36-L41, and https://github.com/privacypass/challenge-bypass-extension/blob/master/src/ext/issuance.js#L240-L242. If you specify your original commitments as part of the "HC" config with version "1.0", and then just remove the `getConfigName() === "CF"` check, then you should be good to go.

Also it seems that the linter is a bit sad now.

alxdavids:
Just to clarify, if you make the changes that I've stated above, then you won't need the extra logic about cancelling the XHR. The commitments will just be retrieved directly from the config.

durch:
@alxdavids thank you for the quick reply! We'll reopen against master shortly, with changes added in :)

PR #133, opened 2019-10-24
Use commitments from config, fixes to issue and redeem for no-reload

durch:
@alxdavids Resubmitting our latest compatibility fixes, as per your suggestion we're using commitments from `config.js`, also fixing some issues we've found with the `no-reload` method.

Feedback is very welcome!

### DIFF
--- a/src/ext/background.js
+++ b/src/ext/background.js
@@ -460,7 +460,7 @@ function beforeRequest(details, url) {
 
     // actually send the token signing request via xhr and return the xhr object
     const xhr = sendXhrSignReq(xhrInfo, url, details.tabId);
-    
+
     /** In the no-reload paradigm the issuance request is sent along side the original solve request. Requests are reconciled on the backend.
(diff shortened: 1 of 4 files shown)


src/ext/background.js
```diff
@@ -460,6 +460,9 @@ function beforeRequest(details, url) {
 
     // actually send the token signing request via xhr and return the xhr object
     const xhr = sendXhrSignReq(xhrInfo, url, details.tabId);
+    if (xhrInfo.cancel) {
```
alxdavids: I'm not sure exactly what this is being used to do?

src/ext/issuance.js
```diff
@@ -62,15 +62,16 @@ function signReqCF(url) {
 /**
  * hCaptcha issuance request
  * @param {URL} url
+ * @param {DETAILS} details
```
alxdavids: nit: `{Object}`

durch:
@alxdavids thank you for the quick turnaround, fixed that silly definition error, also added a brief description of the issue workflow. I'd be happy to hop on a call and go over it in more detail :)

PR #136, opened 2019-10-27
Remove captcha-domain

durch:
@alxdavids while we were testing latest version on our production stack we realized that the `captcha-domain` config setting is completely breaking our redeem flow. Our production URLs are actually on the captcha domain itself, which of course make redeem impossible. Unfortunately we did not put two and two together until now.

It would we incredibly helpful if you could merge this and push out a new version of the extension ASAP.

Thank you!


durch:
Thank you @alxdavids, I already have the tests in mind, will carve out chunk of time an PR those in as well.

PR #142, opened 2019-11-27
`Get more passes` url based on activeConfig

durch:
Sets `Get more passes` url based on the currently active config. Still a WIP while we settle on the URL to use for HC config. 

Additionally I've noticed an issue with CF issuance, namely I can't seem to get any tokens from `captcha.website`. While this is unrelated to the changes above we have been getting complaints along the lines of what I'm seeing. So I'll spend some time looking into that as well.

### DIFF
--- a/src/ext/background.js
+++ b/src/ext/background.js
@@ -93,7 +93,6 @@ let optEndpoints = () => activeConfig()["opt-endpoints"];
 let emptyRespHeaders = () => activeConfig()["spend-action"]["empty-resp-headers"];
 let storedCommitments = () => activeConfig()["commitments"];
 
-
 /**
  * Allows access to get-more-passes-url from UpdatePopup
  * @return {string}
--- a/src/ext/background.js
+++ b/src/ext/background.js
@@ -34,6 +34,7 @@
 /* exported getConfigName */
 /* exported storedCommitments */
 /* exported requestIdentifiers */
+/* exported getMorePassesUrl */
 
 "use strict";
 
@@ -91,6 +92,15 @@ let tokensPerRequest = () => activeConfig()["issue-action"]["tokens-per-request"
 let optEndpoints = () => activeConfig()["opt-endpoints"];
 let emptyRespHeaders = () => activeConfig()["spend-action"]["empty-resp-headers"];
 let storedCommitments = () => activeConfig()["commitments"];
+
+
+/**
+ * Allows access to get-more-passes-url from UpdatePopup
+ * @return {string}
+ */
+function getMorePassesUrl() {
+    return activeConfig()["get-more-passes-url"];
+}
 
 /* Config variables that are reset in setConfig() depending on the header value that is received (see config.js) */
 initECSettings(h2cParams());
--- a/src/ext/config.js
+++ b/src/ext/config.js
@@ -72,6 +72,7 @@ function exampleConfig() {
             "clearance-cookie": "", // name of clearance cookies for checking (cookies that are optionally acquired after redemption occurs)
         },
         "captcha-domain": "", // optional domain for acquiring tokens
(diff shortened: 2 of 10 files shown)


durch:
We have the URL in, ready for review

src/ext/background.js
```diff
@@ -90,6 +91,15 @@ let optEndpoints = () => activeConfig()["opt-endpoints"];
 let emptyRespHeaders = () => activeConfig()["spend-action"]["empty-resp-headers"];
 let storedCommitments = () => activeConfig()["commitments"];
 
+
+/**
+ * Allows access to get-more-passes-url from UpdatePopup
+ * @return {string}
+ */
+function getMorePassesUrl() {
```
alxdavids: Nit: would it be possible to style this in the same as the other configuration functions above?

src/ext/config.js
```diff
@@ -107,6 +108,7 @@ function PPConfigs() {
     cfConfig["issue-action"]["tokens-per-request"] = 30;
     cfConfig.cookies["clearance-cookie"] = "cf_clearance";
     cfConfig["captcha-domain"] = "captcha.website";
+    cfConfig["get-more-passes-url"] = "https://captcha.website";
```
alxdavids: Nit: could we set `"captcha.website"` as a variable and then use that in both `cfConfig["captcha-domain"]` and `cfConfig["get-more-passes-url"]`. Alternatively, there may be a nice way of just removing the `captcha-domain` key as it is only used by CF and this seems to duplicate the usage somewhat.

test/config.test.js
```diff
@@ -0,0 +1,25 @@
...
+
+each(PPConfigs().filter((config) => config.id > 0).map((config) => [config.id]))
+    .describe("CONFIG_ID = %i", (configId) => {
+        beforeEach(() => {
+            workflow.__set__("CONFIG_ID", configId);
+        });
+
+        test("ensure `get-more-passes-url` is a valid URL", () => {
+            new URL(activeConfig()['get-more-passes-url']);
+        })
```
alxdavids: Maybe we could add a test to ensure that the URL is fixed to the correct value for both configs? I doubt the CF value may change, but not sure if this is also the case for HC.

src/ext/config.js
durch: I'll give it some thought, I think `captcha-website` is also used when checking if we should redeem, so it serves another purpose

src/ext/background.js
durch: I had issues with using it in the `popup` context, thats why its using the canonical form, I can spend some time figuring it out, its bugging me as well :)

src/ext/config.js
durch: We also use it to clear cookies on the captcha domain, so I'd leave it in.

durch:
Rebased with master to resolve conflicts

src/ext/config.js
alxdavids: Okay, would it be possible to set `"captcha.website"` as a variable in this file and then just use it in both cfConfig settings?

alxdavids:
@armfazh if you also have time to review this, that would be great

test/.eslintrc.json
```diff
@@ -63,7 +63,8 @@
         "validateRespMock": true,
         "workersG": true,
         "workersH": true,
-        "workflowSet": true
+        "workflowSet": true,
+        "getMorePAssesUrl": true
```
armfazh: Nit: this list must be sorted

durch: @armfazh done!

alirezajavani:
https://github.com/privacypass/challenge-bypass-extension/pull/142#issuecomment-559216074

PR #143, opened 2019-11-27
Fixes #141

### LINKED ISSUE #141: Always has "0 passes".
I always have 0 passes. Clicking "Get More Passes" takes me to captcha website and then
* I cannot get pass through captcha. Many tries gives nothing.
* Sometimes I am verified as human and then I go to privacypass.github.io. Although verification was successful I still have 0 passes.

I am asked about verification almost on every page now. Even if I verify my self I am asked again in two minutes.

alxdavids:
Cloudflare has changed the URL used for sending CAPTCHA solutions.
- This meant that the extension never intercepted the new requests
- This change modifies the expected URL for issuance requests

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -176,6 +176,12 @@ Cloudflare, there is also a server-side upper bound of 100 tokens for each
 signing request. We recommend that this is enforced to prevent unlimited numbers
 of tokens being signed.
 
+#### config\["issue-action"\]\["request-identifiers"\]
+
+An object containing optional identifying characteristics of requests. Requests
+identified using these characteristics can be augmented with data required for
+the issuance portion of the Privacy Pass protocol.
+
 ### config\["cookies"\]
 
 A JSON struct of configuration related to cookie management.
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -182,6 +182,13 @@ An object containing optional identifying characteristics of requests. Requests
 identified using these characteristics can be augmented with data required for
 the issuance portion of the Privacy Pass protocol.
 
+For example, the CF config (id=1) uses the following parameters:
+
+- "query-param": for identifying requests based on the URL query parameters
+- "body-param": for identifying requests based on HTTP request body parameters
+- "post-processed": for identifying requests that have already been processed by
+the WebRequest API
+
 ### config\["cookies"\]
 
 A JSON struct of configuration related to cookie management.
--- a/src/ext/config.js
+++ b/src/ext/config.js
@@ -106,7 +106,7 @@ function PPConfigs() {
     cfConfig["spending-restrictions"]["status-code"] = [403];
     cfConfig["spend-action"]["redeem-method"] = "reload";
     cfConfig["issue-action"]["tokens-per-request"] = 30;
-    cfConfig["issue-action"]["request-identifiers"] = {"query-param": "__cf_chl_captcha_tk__", "body-param": "g-recaptcha-response", "post-processed": "&captcha-bypass=true"};
+    cfConfig["issue-action"]["request-identifiers"] = {"query-param": "__cf_chl_captcha_tk__", "body-param": "g-recaptcha-response", "post-processed": "captcha-bypass"};
     cfConfig.cookies["clearance-cookie"] = "cf_clearance";
     cfConfig["captcha-domain"] = "captcha.website";
     cfConfig["send-h2c-params"] = true;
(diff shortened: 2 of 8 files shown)


src/ext/issuance.js
```diff
@@ -33,28 +33,31 @@ const COMMITMENT_URL = "https://raw.githubusercontent.com/privacypass/ec-commitm
...
  * @return {XMLHttpRequest} XHR info for asynchronous token issuance
  */
-function signReqCF(url) {
+function signReqCF(url, details) {
     const reqUrl = url.href;
-    const manualChallenge = reqUrl.includes("manual_challenge");
-    const captchaResp = reqUrl.includes("g-recaptcha-response");
-    const alreadyProcessed = reqUrl.includes("&captcha-bypass=true");
+    const captchaResp = reqUrl.includes(requestIdentifiers()["query-param"]);
+    const alreadyProcessed = reqUrl.includes(requestIdentifiers()["post-processed"]);
```
Lekensteyn: If suggest to use the [searchParams][2] field instead of performing substring parsing.

```js
const captchaResp = url.searchParams.get(requestIdentifiers()["query-param"]);
// be sure to drop the '&' from the config.js
const alreadyProcessed = url.searchParams.has(requestIdentifiers()["post-processed"]);
// alternatively:
const alreadyProcessed = url.searchParams.get(requestIdentifiers()["post-processed"]) === 'true';
```

 [2]: https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams

src/ext/issuance.js
```diff
@@ -33,28 +33,31 @@ const COMMITMENT_URL = "https://raw.githubusercontent.com/privacypass/ec-commitm
...
     // Generate tokens and create a JSON request for signing
     const tokens = GenerateNewTokens(tokensPerRequest());
-    const request = BuildIssueRequest(tokens);
+    const btRequest = BuildIssueRequest(tokens);
 
     // Tag the URL of the new request to prevent an infinite loop (see above)
     const newUrl = markSignUrl(reqUrl);
     // Construct info for xhr signing request
-    const xhrInfo = {newUrl: newUrl, requestBody: "blinded-tokens=" + request, tokens: tokens};
+    const bodyCaptcha = `${captchaKey}=${bodyParam}`;
```
Lekensteyn: In case the response contains garbage, `${encodeURIComponent(bodyParam)}` to prevent injection?

docs/CONFIG.md
```diff
@@ -176,6 +176,12 @@ Cloudflare, there is also a server-side upper bound of 100 tokens for each
 signing request. We recommend that this is enforced to prevent unlimited numbers
 of tokens being signed.
 
+#### config\["issue-action"\]\["request-identifiers"\]
+
+An object containing optional identifying characteristics of requests. Requests
+identified using these characteristics can be augmented with data required for
+the issuance portion of the Privacy Pass protocol.
```
Lekensteyn: This documentation is incomplete, someone would have to look at the source code to understand what values to put here.

src/ext/issuance.js
```diff
@@ -38,8 +38,8 @@ const COMMITMENT_URL = "https://raw.githubusercontent.com/privacypass/ec-commitm
  */
 function signReqCF(url, details) {
     const reqUrl = url.href;
-    const captchaResp = reqUrl.includes(requestIdentifiers()["query-param"]);
-    const alreadyProcessed = reqUrl.includes(requestIdentifiers()["post-processed"]);
+    const captchaResp = url.searchParams.has(requestIdentifiers()["query-param"]);
```
Lekensteyn: should be get instead of has

migueldemoura:
> This might need extra care to handle metals that are not yet updated.

Indeed, maybe keep both detection options?

src/ext/issuance.js
alxdavids: I think `has` is fine because we don't modify this part of the URL and it is used to send the subsequent XHR request, no?

src/ext/config.js
```diff
@@ -105,6 +106,7 @@ function PPConfigs() {
     cfConfig["spending-restrictions"]["status-code"] = [403];
     cfConfig["spend-action"]["redeem-method"] = "reload";
     cfConfig["issue-action"]["tokens-per-request"] = 30;
+    cfConfig["issue-action"]["request-identifiers"] = {"query-param": "__cf_chl_captcha_tk__", "body-param": "g-recaptcha-response", "post-processed": "captcha-bypass"};
```
armfazh: Some public documentation about these paths is available?

migueldemoura: There's no documentation for the query parameter. Not sure about the other ones.

alxdavids: +1, added some more documentation

PR #150, opened 2020-03-26
Configuration updates

alxdavids:
- Remove tests that depend on real entries (expiry date issues)
- Add extra captcha response fields to cf config
- Prevent conflicts from config changes from overlayed issuers
- Small changes to tests to accommodate fixes


alxdavids:
@migueldemoura in case you want to take a look

migueldemoura:
Cloudflare config changes look good 👍

PR #154, opened 2020-04-01
Issue tokens for all configs

alxdavids:
There are conflicts between the CF and HC configs. When a `cf-chl-bypass: id` response header is received it used to change the active config to the config with ID set to `id`. This causes problems as HC CAPTCHAs would override the active config on CF pages. Issuance requests were only sent for the active config and so they were never sent for CF.

This change stores all response IDs of this form and queues future issuances. At some point, issuance requests are sent on all requests that correspond to a previous ID that was received.


src/ext/background.js
```diff
@@ -690,3 +694,13 @@ function setConfig(val) {
...
 }
+
+/**
+ * Checks that the URL initiating the config switch isn't specifically
+ * ignored
+ * @param {string} url URL string to check
+ * @return {boolean} true if ignored
+ */
+function ignoredUrl(url) {
+    return switchConfigIgnoreUrls().some((u) => url.includes(u));
```
Lekensteyn, line 705: This would also match `http://not.hcaptcha.com/getcaptcha`. or `https://example.com/not/hcaptcha.com/getcaptcha-fake`. Is that a problem?

docs/CONFIG.md
```diff
@@ -262,5 +262,11 @@ full).
 
 ### config\["send-h2c-params"\]
 
-A boolean that determines whether the contents of config\["h2c-params"\] should
+A boolean that determines whether the contents of `config\["h2c-params"\]` should
```
Lekensteyn, line 265: Shouldn't the backslashes be removed if you use backticks?

alxdavids:
Updated PR, see new description. Also rebased on UI changes.

alxdavids:
cc @Lekensteyn, @armfazh

armfazh:
For some reason, after the first issue of tokens, any other request for more tokens causes "rate limited or network error" appears in the hCaptcha widget. So no more tokens are issued.

alxdavids:
Yeah I've noticed this, it seemed to me like it could be an hCaptcha server-side thing? It worked fine when I tried the CAPTCHA on hcaptcha.com or https://www.hcaptcha.com/privacy-pass.



---- On Thu, 02 Apr 2020 19:02:52 +0100 Armando Faz <notifications@github.com> wrote ----




For some reason, after the first issue of tokens, any other request for more tokens causes "rate limited or network error" appears in the hCaptcha widget. So no more tokens are issued.

Issue #153, opened 2020-03-31
Improve extension UI for different configurations


alxdavids:
Currently, the extension only displays information for the currently active configuration. It would be better to display the token counts (and "Get more passes" link) for all available configurations (currently CF & hCaptcha).

PR #155, opened 2020-04-01
Re-design UI to show both configurations (#153)

alxdavids:
Summary:

- Show both configurations at the same time (demarcate which is active)
- Remove "Get More Passes" button
- Replace with `mouseover` characteristics to get more passes.

Images showing proposed changes (feedback welcome!):

Old UI:

image

New UI:

image
image

### DIFF
--- a/addon/popup/bc-plugin.js
+++ b/addon/popup/bc-plugin.js
@@ -15,40 +15,37 @@ if (background) {
 }
 
 function UpdatePopup() {
-    let tokLen = 0
     if (background) {
-        tokLen = background.countStoredTokens();
-        handleResponse(tokLen, background.getMorePassesUrl());
+        let configTokLens = background.getTokenNumbersForAllConfigs();
+        handleResponse(configTokLens);
     } else {
-        let send = browser.runtime.sendMessage({
-            tokLen: true
-        });
-        send.then(handleResponse);
+        browser.runtime.sendMessage({
+            tokLen: true,
+        }).then(handleResponse);
     }
 }
 
-function handleResponse(tokLen, url) {
+function handleResponse(configTokLens) {
     // Replace the count displayed in the popup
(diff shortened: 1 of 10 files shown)


addon/popup/bc-plugin.js
```diff
@@ -91,31 +87,51 @@ function outputRedemption(v, s1, s2) {
...
-    newCount.appendChild(document.createTextNode(tokLen));
-    document.getElementById("stored").appendChild(newCount);
+        stored.onmouseover = () => {
+            const passesText = document.getElementById(`passtext-${ele.id}`);
+            passesText.textContent = "Get more passes!";
+        };
+        stored.onmouseleave = () => {
+            const passesText = document.getElementById(`passtext-${ele.id}`);
+            passesText.textContent = `${ele.name} ${active}`;
+        }
```
Lekensteyn, line 131: Missing semicolon.

addon/popup/bc-plugin.js
```diff
@@ -15,40 +15,37 @@ if (background) {
...
+                UpdatePopup();
+            })
         }
     });
 
     // this allows the client to generate a redemption token for CF API.
     document.getElementById("redeem").addEventListener("click", () => {
         if (background) {
-            tokLen = background.countStoredTokens();
+            tokLen = background.countStoredTokens(background.getConfigId());
```
armfazh, line 48: The extension shows an error when I clicked on "Log redemption console"

>  Uncaught TypeError: background.getConfigId is not a function

armfazh:
How do one can toggle between configurations?

alxdavids:
@armfazh The redemption button only corresponds to the CF configuration as it is only intended to be used with the redemption API at privacypass.cloudflare.com/api/redeem. I've reworded the text to make it a bit clearer.

PR #160, opened 2020-04-15
Allow configurations to be patched externally

### LINKED ISSUE #152: Move configuration to separate repository
Currently, when CF or hCaptcha changes their configuration we also have to release a new version of the extension. It would be preferable to move this configuration to a separate repository (similarly to how commitments are currently stored elsewhere: https://github.com/privacypass/ec-commitments). The configuration should be retrieved over HTTP when the browser first loads. We can also just process configuration patches from the separate repository and display a default config natively in the extension.

alxdavids:
Fixes #152. Allows configs to be patched externally to prevent requiring an extension release for configuration update.

- config patches can be pushed to privacypass/ec-commitments
- patches are applied when extension starts
- patches must be verified by same public key used to sign commitments


src/ext/config.js
```diff
@@ -157,8 +176,112 @@ function PPConfigs() {
...
+        }
+
+        const config = getConfigForId(cfgId);
+        Object.keys(patches).forEach((key) => {
+            if (!PATCHABLE_KEYS.includes(key)) {
+                // do not process patches for non-patchable fields
+                console.warn(`Patches for ${key} are not permitted`);
+                return;
+            }
+            const patch = patches[key];
```
armfazh, line 237:
```suggestion
            const patchValue = patches[key];
```

docs/CONFIG.md
```diff
@@ -269,3 +269,80 @@ full).
...
+
+## Patching configurations
+
+Configurations can be patched by adding patches to the configuration
+available at <https://github.com/privacypass/ec-commitments>. This
+allows external configuration changes to be made without requiring an
+extension update. A patch blob must be signed with the secret key
+corresponding to the public available at `config["comm-vk"]`.
+
+Currently, the configuration available at the URL above looks something
```
armfazh, line 281:
```suggestion
Previously, the configuration available at the URL above looks something
```

docs/CONFIG.md
```diff
@@ -231,41 +243,118 @@ A JSON struct of configuration related to error codes.
...
-If it is set to "increment" then we use the hash-and-increment method
-that will be deprecated moving forward. If it is set to "swu", then we use the
-affine version of the SWU algorithm implemented in h2c.js (see
-\[HASH_TO_CURVE.md\](docs/HASH_TO_CURVE.md) for a description of the algorithm in
-full).
+The hash-to-curve method that is used for moving field elements to curve
+points. If it is set to "increment" then we use the hash-and-increment
+method that will be deprecated moving forward. If it is set to "swu",
+then we use the affine version of the SWU algorithm implemented in
+h2c.js (see [HASH_TO_CURVE.md](docs/HASH_TO_CURVE.md) for a
```
Lekensteyn, line 277: I think that you have to drop `docs/` from this link

alxdavids:
> Looks reasonable. Do you require users to upgrade to the latest extension version? If not, then you may end up with a config that breaks older extension versions.

Users will need to upgrade their extension once I have done a release (after this change) before patches will be accepted.

Lekensteyn:
> > Looks reasonable. Do you require users to upgrade to the latest extension version? If not, then you may end up with a config that breaks older extension versions.
> 
> Users will need to upgrade their extension once I have done a release (after this change) before patches will be accepted.

Right, I was thinking about future changes. You might end up with a config patch that requires a new extension version.

If I am not mistaken, this init code is executed only once in the background page, so unless the user reloads their extension or updates the browser, it might not have the desired update frequency. To run something periodically, have a look at the [alarms](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/alarms) API.

alxdavids:
> Right, I was thinking about future changes. You might end up with a config patch that requires a new extension version.

That is true, perhaps the best approach would be to include an extra key into the `patches` structure which specifies the minimum version that the patches should be applied for? Like:

```json
{
  "patches":
    "2": {
      ...
    },
    "3": {
      ...
    },
    ... etc ...
}
```

> If I am not mistaken, this init code is executed only once in the background page, so unless the user reloads their extension or restarts the browser, it might not have the desired update frequency. To run something periodically, have a look at the [alarms](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/alarms) API.

I'm not *super* worried about patches not being applied if the user does not restart the browser. However, I can see the benefit of doing periodic updates as well so I will see what I can do.

Lekensteyn:
What about turning `patches` into an array? While doing so, one could also move metadata such as the signature from the config structure.

```json
{
    "patches": [
        {
            "min-version": "3.14",
            "config": {
                "captcha-domain": "example.com"
            },
            "sig": "<signature value 1>"
        },
        {
            "min-version": "3.141",
            "config": {
                "captcha-domain": "example.net"
            },
            "sig": "<signature value 2>"
        }
    ]
}
```

> I'm not super worried about patches not being applied if the user does not restart the browser.

I originally though that the change would affect the commitments refresh as well, but it appears that the `createVerificationXHR` continues to refresh these when a version is missing. Maybe it would make sense to have one point where both the commitments and the patches are downloaded, cached and applied?

alxdavids:
@Lekensteyn I agree that the config structure that you've proposed is better than what I was aiming at. I've pushed an update that configures the patches in this way, let me know what you think.

> Maybe it would make sense to have one point where both the commitments and the patches are downloaded, cached and applied?

I think this would be worthwhile but I think it's sufficiently separate from this change that I would like to address it in a future change.

joshuamorris23:
@alxdavids

Makefile
```diff
@@ -1,14 +1,15 @@
 SOURCES= src/crypto/keccak/keccak.js \
 		 src/crypto/local.js         \
+         src/ext/config.js           \
```
maxius122-bit, line 3: jason serrano

maxius122-bit: jason serrano

addon/popup/bc-plugin.js
```diff
@@ -8,10 +8,18 @@
 let background = chrome.extension.getBackgroundPage();
 if (background) {
     background.UpdateCallback = UpdatePopup;
+    // update version text
+    document.getElementById("header-text").innerHTML = `<h1>Privacy Pass<br><font size="1">Version ${background.extVersion()}</font></h1>`;
```
maxius122-bit, line 12: jason serrano

addon/popup/bc-plugin.js
```diff
@@ -8,10 +8,18 @@
...
 } else {
     browser.runtime.sendMessage({
         callback: UpdatePopup,
     });
+    // update version text
+    browser.runtime.sendMessage({
+        version: true,
+    }).then((extVersion) => {
```
maxius122-bit, line 20: jason serrano

PR #169, opened 2020-05-08
Bug fix: Montgomery powermod

mavroudisv:
Montgomery powermod is not always possible. In those cases, "montpowermod" returns 'false' which results in an error. Instead, it's best to call powermod which uses "montpowermod" when possible and falls back to classic powermod otherwise.


alxdavids:
Seems fine to me. Merging.

PR #172, opened 2020-05-09
WIP: Support for P384 and P521

mavroudisv:
Working in progress for supporting curves P384 and P521. Note that some of the tests for P384 fail.


alxdavids:
Hey @mavroudisv, where did you get the expected values for the P-384 tests from? As you mentioned previously, it seems that the discrepancy is coming from the computation of `e` during SWU. I'll need to check exactly what this means.

Apart from that, the rest of the change looks pretty sane to me. I think the only thing I would say is that it may be worth trying to combine all the curve tests into a single file as they repeat a lot of very similar code. You can declare tests within a for loop that iterates over all of the curves. Then you should be able to fix the constant values (e.g. expectations) depending on the curve that is being used.

mavroudisv:
Yeah, that's the other issue we discussed previously; the values were generated by the same code they are tested on. Is there any other project implementing those algorithms that we could somehow use/abuse to generate test vectors?

The test that fails (where the X coordinate is expected to be 0 but is not) could just be the only case that it's clearly buggy. I don't have a lot of trust in the correctness of the rest of the test triplets (scalar, X, Y).

You're right that the curve tests can be combined. Having them separately was just for convenience to speed up development as I would run the tests only for the curve I was working on (even though jest probably has a way to specify input arguments that would restrict the tests to only one curve).

alxdavids:
> The test that fails (where the X coordinate is expected to be 0 but is not) could just be the only case that it's clearly buggy. I don't have a lot of trust in the correctness of the rest of the test triplets (scalar, X, Y).

I don't think this necessarily *must* be a bug, it may just be that the SWU algorithm acts slightly different for P384. I think determining the expected values from this code is not too much of a problem for now. I just wondered why there were errors in these tests, if the expected values were derived from the code itself?

mavroudisv:
Fair enough, I was convinced that it must be a mistake (the fact that it doesn't return 0 in X) so I didn't update the expected values. Happy to do so.

Nubovik01:
whaaaaaaaaaat

all420:
> addon/manifest.jsonhttps://github.com/privacypass/challenge-bypass-extension/communityhttps://github.com/privacypass/challenge-bypass-extension/blob/f6f71a8557339cd702057ee6c25db1120ddd4dda/test/issuance.test.jsfa64de1keep me postedhttps://github.com/privacypass/challenge-bypass-extension/communityfa64de1

all420:
https://github.com/privacypass/challenge-bypass-extension/blob/f6f71a8557339cd702057ee6c25db1120ddd4dda/test/issuance.test.js

thibmeu:
This is likely a late comment, but thanks for working through an initial implementation.
Since this PR was first opened, the Privacy Pass protocol has evolved within IETF, and P251 is not included in [the current draft](https://www.ietf.org/archive/id/draft-ietf-privacypass-protocol-16.html#name-token-type-registry-updates).

Existing implementation for the standard protocol are tracked in a dedicated repository: https://github.com/ietf-wg-privacypass/base-drafts#existing-implementations

I'm closing this PR

PR #173, opened 2020-05-16
Switch submodule from git to https protocol

da2x:
Adds protocol layer security (domain name verification) while removing the requirement to have a public-key configured.


alxdavids:
Agree with @Lekensteyn, happy to make the change as long as the commit message is updated.

da2x:
Amended.

alxdavids:
Thanks!

PR #199, opened 2020-10-08
Adding 128x128 icons

armfazh:


alxdavids:
Where will these icons be used? Do you need to add support for them to the manifest?

armfazh:
Good catch, they need be declared in the manifest.
https://developer.chrome.com/extensions/manifest/icons

alxdavids (approved):

Issue #223, opened 2021-03-12
use randomness from webcrypto


armfazh:
https://github.com/privacypass/challenge-bypass-extension/blob/10ef8686b7449f3d01cc5dcc71a4189f8ee88475/src/crypto/local.js#L127

PR #230, opened 2021-05-03
Use webcrypto for randomness (#223)

Jytesh:


src/crypto/local.js
```diff
@@ -122,9 +122,7 @@ function unblindPoint(b, Q) {
  * @return {sjcl.ecc.point}
  */
 function newRandomPoint() {
-    const byteLength = 32;
-    const wordLength = byteLength / 4; // SJCL 4 bytes to a word
-    const random = sjcl.random.randomWords(wordLength, 10); // TODO Use webcrypto instead.
+    const random = crypto.getRandomValues(new Int32Array(8)); // TODO Use webcrypto instead.
```
armfazh, line 125:
```suggestion
    const random = crypto.getRandomValues(new Int32Array(8));
```

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "PrivacyPass",
+  "name": "privacypass",
```
armfazh, line 2: what's the reason for changing this?

armfazh (commented):
@Jytesh , I've added a couple of comments.

package.json
Jytesh: npm was warning me about it violating naming conventions, didn't mean to commit that

Jytesh: You can revert that

armfazh: Didn't found any offfenders in https://docs.npmjs.com/package-name-guidelines
could you please revert the change.
CI is not happy.

Jytesh: image
ok

Jytesh: Also need to make crypto a global to make CI pass

armfazh:
@Jytesh you might need to rebase on top of #234

armfazh (approved):

PR #232, opened 2021-05-06
Removes json non-used dependency.

armfazh:

PR #235, opened 2021-05-11
Bumping up dependencies.

armfazh:
also json dependency is not needed.


Lekensteyn (commented):
The json dependency was already dropped from package.json in git master via https://github.com/privacypass/challenge-bypass-extension/pull/232

Could you rebase this on git master and redo the bump (only if needed)? At that time, the json dependency was already removed from the .lock file as well.

claucece (approved):

Issue #226, opened 2021-04-06
Brave Browser Privacy Pass not work anymore


usafety:
I use Brave Browser and its notw work anymore

AirBashX:
me too,There is no longer role in the Chrome89 version.

BTCAlchemist:
Same issue for me. For example, when I go to etherscan.io, it prompts me to complete hCaptcha. When I click Privacy Pass to activate the browser extension and reload the page, then click the hCaptcha check box, I am still required to solve the hCaptcha puzzle. This happens with Brave shields on and off, and I noticed it sometime last year. Solving this could really save users time and make Privacy Pass worthwhile. I am using Brave browser on Windows 10. @alxdavids Can you help?

![image](https://user-images.githubusercontent.com/56145569/116590447-9fa63300-a8d2-11eb-907d-e5b462894252.png)

7ik1t4:
For me, it's working. May be update it
![image](https://user-images.githubusercontent.com/46179582/116814113-ca171b00-ab74-11eb-9c90-7e605ef46716.png)

usafety:
Same problem

Sent with [ProtonMail](https://protonmail.com) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

BTCAlchemist:
@armfazh @durch Are you able to help?

armfazh:
Using Brave
```
Brave	1.23.75 Chromium: 90.0.4430.93 (Official Build) (64-bit)
Revision	4df112c29cfe9a2c69b14195c0275faed4e997a7-refs/branch-heads/4430@{#1348}
OS	Linux
```

Go to collect passes by clicking in [PP] -> Cloudflare -> Get More Passes.
This prompts [captcha.website](https://captcha.website) and solve the captcha.
With 30 CF and 5 hCaptcha passes, you can go to a website protected by Cloudflare and spend these passes.
You can verify whether a site is behind Cloudflare using [Claire](https://chrome.google.com/webstore/detail/claire/fgbpcgddpmjmamlibbaobboigaijnmkl)
However note that each website could turn off PrivacyPass at its sole discretion.

armfazh:
> BTCAlchemist
> Same issue for me. For example, when I go to etherscan.io, it prompts me to complete hCaptcha. When I click Privacy Pass to activate the browser extension and reload the page, then click the hCaptcha check box, 

[...]

Issue #225, opened 2021-04-04
Privacy Pass not working for current version of browsers


zydjohnHotmail:
Before filing a new bug, verify if your request is already being tracked by another issue here: https://github.com/privacypass/challenge-bypass-extension/issues.

--- 

If you believe that this is a new bug, please proceed to create an issue. The issue will be investigated after you have filled in the following information.

**Describe the bug**
A clear and concise description of the bug.

**To Reproduce**
1. Go to 'https://buff.bet/'
2. Click on 'I am human'
3. Scroll down to picture to solve the captcha.

**Expected behavior**
I should get 5 passes, which I can use to avoid solve captcha for 5 more times.

**System (please complete the following information):**
 - OS: Microsoft Windows 10 (Version 10.0.19042.870)
 - Cloudflare tokens or hCaptcha tokens?
 - hCaptcha tokens
 - Browser [e.g. chrome, firefox]
 - Both Chrome and Firefox.
 - Browser Version [e.g. 79, 80, ]
 - Chrome Version 89.0.4389.114 (64-bit)
 - Firefox Version 87.0 (64-bit)
 - Privacy Pass Version [e.g. 2.0.4, 2.0.5 ]
 - 2.0.8
 - Did you install Privacy Pass from this repository or from the browser store?
 - Install from the browser store

**Additional context**

There is one web site, which migrated to use hCaptcha at the end of year 2020.  At the first a few months, I installed Privacy Pass extension for Chrome, it worked. 
However, until rather recently, with the latest version of Chrome Version 89.0.4389.114 (Official Build) (64-bit) and Fire-Fox Version 87.0 (64-bit).
Privacy Pass extension version 2.0.8 stops working any more.
Usually, the first time, when I visit the underlying web site, I have to solve the captcha, then I can get 5 passes, then after exactly 2 hours, the first pass expired; I can simply click on “I am human” button to extend captcha-free time for another 2 hours.
But with the current browsers’ version and current version of privacy pass extension, after 2 hours, I have to manually solve the captcha again.
Please check to verify if you can reproduce the issue.
By the way, if I want to change the extension, now I can get 2.7.0.zip file, so after 2 hours, I want to  automatically click on “I am human” button.  Let me know if it is possible, if yes, give me some general idea how, and provide some guide or reference url.
Thanks,

CMDR-Dantae:
[...]

PR #237, opened 2021-05-18
Ignore query strings in the hcaptcha spend/issue urls

### LINKED ISSUE #224: hCaptcha passes not working
**Describe the bug**
As seen in the screenshot, I do have hCaptcha passes but the the [webpage](https://forum.xda-developers.com) is still asking me for captcha.
image

**To Reproduce**
1. Open https://forum.xda-developers.com

**Expected behavior**
I should be able to visit the website without verifying the captcha if I have passes for the captcha.

**System (please complete the following information):**
 - OS: Microsoft Windows 10 (Version 10.0.19042.867)
 - hCaptcha tokens
 - Browser: Firefox
 - Browser Version: 89.0a1 (2021-04-03) (64-bit)
 - Privacy Pass Version: 2.0.8
 - Add-on installed from Mozilla's add-on store

ppopth:
This resolved #224 #225 #226

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -120,7 +120,8 @@ A JSON struct of configuration related to redeeming tokens.
 #### config\["spend-action"\]\["urls"\]
 
 URLs that activate WebRequest listeners for redemption, "`<all_urls>`"
-corresponds to matching all possible URLs.
+corresponds to matching all possible URLs. All the GET query strings will
+be ignored.
 
 #### config\["spend-action"\]\["redeem-method"\]
 
@@ -165,7 +166,8 @@ A JSON struct of configuration related to issuing tokens.
 #### config\["issue-action"\]\["urls"\]
 
 URLs that activate WebRequest listeners for issuance, "`<all_urls>`"
-corresponds to matching all possible URLs.
+corresponds to matching all possible URLs. All the GET query strings will
+be ignored.
 
 #### config\["issue-action"\]\["sign-reload"\]
 
--- a/src/ext/utils.js
+++ b/src/ext/utils.js
@@ -43,6 +43,7 @@ function patternToRegExp(pattern) {
     reString += "(:\\d+)?";
     reString += "\\/";
     reString += path.replace("*", ".*");
+    reString += "(\\?.*)?";
     reString += "$";
 
     return RegExp(reString);


armfazh (approved):
it works with hcaptcha.

src/ext/config.js
```diff
@@ -179,8 +180,16 @@ function hcBaseConfig() {
...
-    hcConfig["spend-action"]["urls"] = ["https://*.hcaptcha.com/getcaptcha", "https://*.hmt.ai/getcaptcha", "http://127.0.0.1/getcaptcha"];
-    hcConfig["issue-action"]["urls"] = ["https://*.hcaptcha.com/checkcaptcha/*", "https://*.hmt.ai/checkcaptcha/*", "http://127.0.0.1/checkcaptcha/*"];
+    hcConfig["spend-action"]["urls"] = [
+        /^https:\/\/(.*\.)?hcaptcha\.com\/getcaptcha(\?.*)?$/,
+        /^https:\/\/(.*\.)?hmt\.ai\/getcaptcha(\?.*)?$/,
+        /^https:\/\/127\.0\.0\.1\/getcaptcha(\?.*)?$/,
+    ];
+    hcConfig["issue-action"]["urls"] = [
+        /^https:\/\/(.*\.)?hcaptcha\.com\/checkcaptcha(\/.*)?$/,
+        /^https:\/\/(.*\.)?hmt\.ai\/checkcaptcha(\/.*)?$/,
```
Lekensteyn, line 190: `.*` might also match `https://example.com/.hmt.ai/checkcaptcha`, let's try to avoid that with this tweak:
```suggestion
        /^https:\/\/([^/]*\.)?hmt\.ai\/checkcaptcha(\/.*)?$/,
```

This applies to all previous regexes as well. Previously wildcards were replaced by `[^\\/]*` via utils.js.

Lekensteyn (changes requested):
Looks reasonable except two things:

* The regex is too broad
* docs/CONFIG.md needs to be updated since its meaning has changed.

I wonder whether we should actually add the full flexibility of regexes to the config format? As alternative, the query string could be stripped, or multiple patterns could be added.

The advantage of that is that the config change does not have to change. A regex is not a valid type in JSON, so changing from a string to a RegExp object can be somewhat problematic.

Lekensteyn (approved):
LGTM Thanks!

PR #242, opened 2021-07-12
Reduce tokens per request to 5

ppopth:


armfazh (approved):

ValZapod:
WTF. You can do that? Will clouflare accept 100 tokens then too?

PR #251, opened 2021-09-21
Updating several dependencies with vulnerabilites

armfazh:


Lekensteyn (approved):
I didn't look closely at the yarn.lock changes, but the CI changes look fine.
I would probably keep the `sjcl` target separate to speed up builds when developing the extension itself.

PR #252, opened 2021-09-21
Adding github actions for running tests

armfazh:


armfazh:
@dependabot rebase

Lekensteyn (approved):
This change is also in https://github.com/privacypass/challenge-bypass-extension/pull/251, copying feedback from there:

The CI changes look fine.
I would probably keep the sjcl target separate to speed up builds when developing the extension itself.

PR #255, opened 2021-09-22
Removes non-used dependencies

armfazh:


Lekensteyn (approved):
`@babel/core` and `@babel/preset-env` were added in https://github.com/privacypass/challenge-bypass-extension/pull/108 (commit f2886654312f2f493c67c62c925b6914cc68c724) to "prevent some warnings" so I guess it is fine to remove them.

babel-core and all others were added in commit d47fd9d1e017c71e6f93295d71cf3f872105caf7 as part of introducing the jest test framework, so if they are not needed elsewhere it looks good to me.

armfazh:
Well, jest seems to be configured to use Babel.
I think babel is not used, but some configuration is required and am working on that

PR #262, opened 2021-10-11
Remove old Cloudflare captcha endpoint

migueldemoura:
This hasn't been in use for almost 2 years.


armfazh:
Thanks @migueldemoura for the pr, the project has been updated to v3, which does not use that endpoint.

PR #272, opened 2021-10-26
Add Cloudflare unit tests and use callbacks to allow providers to access browser functionalities

ppopth:
There are two commits in the PR. Please look at the commit messages for the detailed information.

### DIFF
--- a/src/background/providers/cloudflare.test.ts
+++ b/src/background/providers/cloudflare.test.ts
@@ -1,0 +1,345 @@
+import { jest } from '@jest/globals';
+import { CloudflareProvider } from './cloudflare';
+import Token from '../token';
+
+export class StorageMock {
+    store: Map<string, string>;
+
+    constructor() {
+        this.store = new Map();
+    }
+
+    getItem(key: string): string | null {
+        return this.store.get(key) ?? null;
+    }
+
+    setItem(key: string, value: string): void {
+        this.store.set(key, value);
+    }
+}
+
+test('getStoredTokens', () => {
+    const storage = new StorageMock();
+    const updateIcon = jest.fn();
+    const navigateUrl = jest.fn();
+
+    const provider = new CloudflareProvider(storage, { updateIcon, navigateUrl });
+    const tokens = [new Token(), new Token()];
+    provider['setStoredTokens'](tokens);
+    const storedTokens = provider['getStoredTokens']();
+    expect(storedTokens.map((token) => token.toString())).toEqual(
+        tokens.map((token) => token.toString()),
+    );
+});
+
+test('setStoredTokens', () => {
+    const storage = new StorageMock();
+    const updateIcon = jest.fn();
+    const navigateUrl = jest.fn();
+
+    const provider = new CloudflareProvider(storage, { updateIcon, navigateUrl });
+    const tokens = [new Token(), new Token()];
+    provider['setStoredTokens'](tokens);
+    const storedTokens = JSON.parse(storage.store.get('tokens')!);
+    expect(storedTokens).toEqual(tokens.map((token) => token.toString()));
+});
+
+test('getBadgeText', () => {
+    const storage = new StorageMock();
+    const updateIcon = jest.fn();
+    const navigateUrl = jest.fn();
+
+    const provider = new CloudflareProvider(storage, { updateIcon, navigateUrl });
+    const tokens = [new Token(), new Token()];
+    provider['setStoredTokens'](tokens);
+    const text = provider['getBadgeText']();
+    expect(text).toBe('2');
+});
+
+/*
+ * The issuance involves handleBeforeRequest listener.
+ * 1. Firstly, the listener check if the request looks like the one that we
+ * should send an issuance request.
+ * 2. If it passes the check, the listener returns the cancel command to
+ * cancel the request. If not, it returns nothing and let the request
+ * continue.
+ * 3. At the same time the listener returns, it calls a private method
+ * "issue" to send an issuance request to the server and the method return
+ * an array of issued tokens.
+ * 4. The listener stored the issued tokens in the storage.
+ * 5. The listener reloads the tab to get the proper web page for the tab.);
+ */
+describe('issuance', () => {
+    describe('handleBeforeRequest', () => {
+        test('valid request', async () => {
+            const storage = new StorageMock();
+            const updateIcon = jest.fn();
+            const navigateUrl = jest.fn();
+
+            const provider = new CloudflareProvider(storage, { updateIcon, navigateUrl });
+            const tokens = [new Token(), new Token(), new Token()];
+            const issue = jest.fn(async () => {
+                return tokens;
+            });
+            provider['issue'] = issue;
+            const url = 'https://captcha.website/?__cf_chl_captcha_tk__=query-param';
+            const details = {
+                method: 'POST',
+                url,
+                requestId: 'xxx',
+                frameId: 1,
+                parentFrameId: 1,
+                tabId: 1,
+                type: 'xmlhttprequest' as chrome.webRequest.ResourceType,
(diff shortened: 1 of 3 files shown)


src/background/providers/cloudflare.test.ts
```diff
@@ -0,0 +1,334 @@
+import { jest } from '@jest/globals';
+import { CloudflareProvider } from './cloudflare';
+import { Storage } from '../storage';
+import Token from '../token';
+
+export class StorageMock {
```
armfazh, line 6: One first change is to convert `Storage` into a interface, so the mock can be instantiated by class implementing `Storage` interface.

src/background/providers/cloudflare.test.ts
```diff
@@ -0,0 +1,334 @@
...
+    store: { [key: string]: string };
+
+    constructor() {
+        this.store = {};
+    }
```
armfazh, line 11:
```suggestion
    store: Map<string, string>;

    constructor() {
        this.store = new Map<string, string>();
    }
```

src/background/providers/cloudflare.test.ts
```diff
@@ -0,0 +1,334 @@
...
+test('getStoredTokens', () => {
+    // @ts-ignore
+    const storage: Storage = new StorageMock();
+    const updateIcon = jest.fn();
+    const navigateUrl = jest.fn();
+
+    const provider = new CloudflareProvider(storage, { updateIcon, navigateUrl });
+    const tokens = [new Token(), new Token()];
+    // @ts-ignore
+    provider.setStoredTokens(tokens);
```
armfazh, line 35: The method is private, so I think is safer to use the index notation method rather than remove the compilers check with `@ts-ignore`.  

```suggestion
    provider["setStoredTokens"](tokens);
```

src/background/providers/cloudflare.test.ts
```diff
@@ -0,0 +1,334 @@
...
+    // @ts-ignore
+    const storedTokens = provider.getStoredTokens(tokens);
```
armfazh, line 37:
```suggestion
    const storedTokens = provider["getStoredTokens"]();
```
As you can see, removing the typescript checker is not good because this method does not receive a parameter, and this was not detected by the checker.

src/background/providers/cloudflare.test.ts
```diff
@@ -0,0 +1,334 @@
...
+    expect(storedTokens.map((token) => token.toString())).toEqual(
+        tokens.map((token) => token.toString()),
+    );
+});
+
+test('setStoredTokens', () => {
+    const mockStorage = new StorageMock();
+    // @ts-ignore
```
armfazh, line 45: The error here is fixable by converting Storage into an interface. Let's do not use `@ts-ignore`.

src/background/providers/cloudflare.test.ts
```diff
@@ -0,0 +1,334 @@
...
+
+                    // Expect the tokens are added.
+                    // @ts-ignore
+                    const storedTokens = provider.getStoredTokens();
+                    expect(storedTokens.map((token) => token.toString())).toEqual(
+                        tokens.map((token) => token.toString()),
+                    );
+                    resolve(null);
+                }),
+            );
```
armfazh, line 126: Here you can use the `await` in front of 
` provider.handleBeforeRequest(details)`
so you can remove the `setTimeout` construction.

```suggestion
            const result = await provider.handleBeforeRequest(details);
            expect(result).toEqual({ cancel: true });

            expect(issue.mock.calls.length).toBe(1);
            expect(issue).toHaveBeenCalledWith(url, {
                ['h-captcha-response']: 'body-param',
            });

            expect(navigateUrl.mock.calls.length).toBe(1);
            expect(navigateUrl).toHaveBeenCalledWith('https://captcha.website/');

            // Expect the tokens are added.
            // @ts-ignore
            const storedTokens = provider.getStoredTokens();
            expect(storedTokens.map((token) => token.toString())).toEqual(
                tokens.map((token) => token.toString()),
            );
```

src/background/providers/cloudflare.test.ts
```diff
@@ -0,0 +1,334 @@
...
+                requestId: 'xxx',
+                frameId: 1,
+                parentFrameId: 1,
+                tabId: 1,
+                type: 'xmlhttprequest' as chrome.webRequest.ResourceType,
+                timeStamp: 1,
+                requestBody: {},
+            };
+            const result = provider.handleBeforeRequest(details);
+            expect(result).toBe(undefined);
```
armfazh, line 151:
```suggestion
            expect(result).toBeUndefined();
```

src/background/providers/cloudflare.test.ts
```diff
@@ -0,0 +1,334 @@
...
+                url: 'https://cloudflare.com/',
+                requestId: 'xxx',
+                frameId: 1,
+                parentFrameId: 1,
+                tabId: 1,
+                type: 'xmlhttprequest' as chrome.webRequest.ResourceType,
+                timeStamp: 1,
+                requestBody: {},
+            };
+            const result = provider.handleBeforeRequest(details);
```
armfazh, line 150:
```suggestion
            const result = await provider.handleBeforeRequest(details);
```

src/background/providers/cloudflare.test.ts
```diff
@@ -0,0 +1,334 @@
...
+            // @ts-ignore
+            const storage: Storage = new StorageMock();
+            const updateIcon = jest.fn();
+            const navigateUrl = jest.fn();
+
+            const provider = new CloudflareProvider(storage, { updateIcon, navigateUrl });
+            const tokens = [new Token(), new Token(), new Token()];
+            // @ts-ignore
+            provider.setStoredTokens(tokens);
+            const details = {
```
armfazh, line 246: just add a comment that tells what makes the call to `handleHeadersReceived` to be invalid, there are many reasons why `handleHeadersReceived` returns nothing.

armfazh (changes requested):
The comments in the commit are really great, better to have them in the testing code too.
Let's remove the `@ts-ignore` as can introduce errors without notice.

ppopth:
> The comments in the commit are really great, better to have them in the testing code too. Let's remove the `@ts-ignore` as can introduce errors without notice.

I use `@ts-ignore` because the tests sometimes need to access/change the private properties

src/background/providers/cloudflare.test.ts
ppopth: `handleBeforeRequest` doesn't return a Promise.

src/background/providers/cloudflare.test.ts
ppopth: ` provider.handleBeforeRequest(details)` doesn't return a Promise, so you can't use `await`

src/background/providers/cloudflare.test.ts
armfazh: but makes code synchronous

src/background/providers/cloudflare.test.ts
armfazh: it doesn't but I think the effect is to synchronize execution. what happened when you tried?

ppopth: I don't think it has any effect. `await` is designed to be used with `Promise` and `handleBeforeRequest` doesn't return any promise. `setTimeout` is used to add a task to the even loop queue which is supposed to be run after any previously pending tasks. https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop#zero_delays for more info

src/background/providers/cloudflare.test.ts
ppopth: Thanks!

armfazh (approved):

armfazh:
Great job, ready to merge

PR #273, opened 2021-10-28
Use callbacks instead of events

ppopth:
For icon updating, we allow the provider to explicitly actively update
the icon using the callback instead of using the event listeners
implemented in the Tab class. By doing this, the provider has the
freedom to update the icon any time it wants.

For tab reloading, we move the logic to the Tab class so that the
provider doesn't have to use the Chrome API directly and allow the
provider to reload the tab using the callback in the same way as in icon
updating.

### DIFF
--- a/src/background/providers/index.ts
+++ b/src/background/providers/index.ts
@@ -3,7 +3,7 @@ export { HcaptchaProvider } from './hcaptcha';
 
 export interface Provider {
     getID(): number;
-    getBadgeText(): string;
+    forceUpdateIcon(): void;
     handleBeforeRequest(
         details: chrome.webRequest.WebRequestBodyDetails,
     ): chrome.webRequest.BlockingResponse | void;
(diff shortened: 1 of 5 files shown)


src/background/providers/index.ts
```diff
@@ -13,4 +13,10 @@ export interface Provider {
     handleBeforeSendHeaders(
         details: chrome.webRequest.WebRequestHeadersDetails,
     ): chrome.webRequest.BlockingResponse | void;
+    handleActivated(): void;
+}
+
+export interface Callbacks {
```
armfazh, line 19: is it possible to move the methods in `Callbacks` to the `Provider` interface?

armfazh (approved):

src/background/providers/index.ts
ppopth: `Callbacks` is supposed to be used in constructors but there is no way to define the constructor signatures in the `Provider` interface.

PR #283, opened 2022-01-16
lots of various improvements..

warren-bank:
both in terms of making the design simpler for adding new providers..
and in terms of adding features (such as re-adding support for the _hCaptcha_ provider).

status:
* builds and runs great

to-dos:
* although I updated the tests, I haven't actually run them yet.. so I'm sure they'll need an hour or two to get everything passing


warren-bank:
[here are releases](https://github.com/warren-bank/fork-crx-PrivacyPass-challenge-bypass/releases) that include unsigned packed extensions.. for browsers that all them to be installed

ex:
* [SRWare Iron](https://www.srware.net/iron/) for Chromium
* Firefix Developer edition

ppopth:
1. Is this still in progress?
2. What did you make .bin a hidden directory?

ppopth:
After all thank you for the patch :)

README.md
```diff
@@ -38,7 +38,7 @@ $ npm ci
 $ npm run build
 ```
 
-After that, the `dist` folder will contain all files required by the extension.
```
ppopth, line 41: Why did you change the name to `PrivacyPass`? I think the name `PrivacyPass` doesn't indicate that the files in the directory are generated files.

ppopth:
I think many directory names and files names are unnecessary too long. For example, `psuedorandom-number-generators`, I think by putting just `prngs`, most people will understand it already

README.md
warren-bank: because when Chrome is used to pack an extension, both the name of the .crx and the .pem are expected to be the same as the input directory (all siblings)..

`.gitignore` allows the .pem to be permanently homed where Chrome expects it to be.. without ever being commit to version control.. and likewise, the generated .crx and .xpi files won't either

this naming convention is associated with the scripts in the `.bin` directory.. if you already have tools that you use to package your extensions, then.. you:
1. won't want/need the `.bin` directory
   * though I find these scripts useful
2. won't want to rename the output directory.. as you tooling will already be configured to use the old path

src/background/providers/cloudflare.ts
```diff
@@ -227,59 +254,71 @@ export class CloudflareProvider implements Provider {
...
         }
 
-        (async () => {
-            // Issue tokens.
-            const tokens = await this.issue(details.url, flattenFormData);
-            // Store tokens.
-            const cached = this.getStoredTokens();
-            this.setStoredTokens(cached.concat(tokens));
+        // delay the request to issue tokens until next tick of the event loop
+        setTimeout(
```
ppopth, line 299: Why do you change to `setTimeout` over here?

warren-bank: because (if memory serves) that's a blocking API method.. and there's no need for the browser to wait for the secondary request to finish; using a next tick allows the triggered code to run asynchronously.

maybe there's a good reason to not do it this way, but it struck me as a better idea while I was touching that portion of code.

src/background/providers/cloudflare.ts
```diff
@@ -1,72 +1,90 @@
...
-const COMMITMENT_URL =
+const COMMITMENT_URL: string =
     'https://raw.githubusercontent.com/privacypass/ec-commitments/master/commitments-p256.json';
 
-const QUALIFIED_QUERY_PARAMS = ['__cf_chl_captcha_tk__', '__cf_chl_managed_tk__'];
-const QUALIFIED_BODY_PARAMS = ['g-recaptcha-response', 'h-captcha-response', 'cf_captcha_kind'];
-
-const CHL_BYPASS_SUPPORT = 'cf-chl-bypass';
-const DEFAULT_ISSUING_HOSTNAME = 'captcha.website';
+const ALL_ISSUING_CRITERIA: {
```
ppopth, line 18: Making the criteria programmable is more flexible than making it configurable. Please read https://blog.cloudflare.com/privacy-pass-v3/ for more detail. We had difficultly by making things configurable in the past because configuration is less flexible.

That is, if we want to change the criteria, it makes more sense to change the code in some function rather change some configuration object.

src/background/providers/cloudflare.ts
warren-bank: actually.. my answer just now was wrong..

```javascript
(async () => {...})()
```

wouldn't block..

I think my reason for adding the timeout in the first place had to do with hCaptcha. I don't entirely understand how their process works.. but it seems (from trial and error) very clear that the intercepted request needs to complete __before__ the triggered request to issue tokens can be made.. otherwise it fails. So.. i don't know if removing this timeout would break hCaptcha or not.. but there's a good chance that it might.

src/background/providers/cloudflare.ts
warren-bank: yes and no..

it would be bad to force the use of a convention by all providers; however, since some tasks are repeated by all providers.. it makes sense to me to provide tooling in the form of static methods that providers have the option to use.

Those tools will of course require formatted input that providers who choose to call these methods will need to supply; that all that these data structures are for.

warren-bank: to make my earlier comment more concrete..

`handleBeforeRequest()` needs to _quickly_ determine whether the intercepted request:
1. solves a capcha belonging to the particular provider
2. is being solved on a domain that issues tokens

these determinations are always based on:
1. the requested url
2. POST data (if present)

I added static methods that allow providers to pass the data from the request as well as a configuration object.. and the result is a simple yes or no.. the request does or doesn't match the filter criteria as configured

If a provider wants to apply specialized logic.. then that's great; simply don't use these methods and roll your own.

src/background/providers/cloudflare.ts
warren-bank: Of course.. these comments about hCaptcha don't apply to the plugin for the Cloudflare provider; the update to Cloudflare was just because I wanted the 2x plugins to keep as nearly identical as possible.. for symmetry.. so any differences in logic can be easily grok'd by diff'ing the two.

warren-bank:
regarding using long/descriptive file and directory names..
* I figure that since the code gets bundled together during the build process and these names are only ever seen by the dev team.. there's no harm in using them
* I'll be the first to admit that I had to google the meaning of `prng` :)
* but.. it's your project.. do as you like

warren-bank:
regarding whether this is still a work in progress..

without getting too philisophical or snarky.. :)
* the resulting extension is stable and works exactly as it should
* I've accomplished my initial goals.. and am satisfied with the result

so.. nope; I'd consider it a release candidate.

of course, we can discuss some of the decisions and make a few tweaks.. but this definitely moves the ball forward.

warren-bank:
One thing that I didn't touch.. but would take a good look at if it was my project.. is the version of the ecmascript in the generated bundle that gets packed into the extension. It's currently set to ES2020.

Personally, I always like to support as many users as possible.. and would probably down level all the way to ES5 (if feasible).

warren-bank:
PS: regarding my earlier comment about needing to test the tests.. I made an additional commit a few hours ago that fixes the tests (mainly to adjust for how to confirm that the code in the new timeouts has run). Now everything passes.

warren-bank:
I just added another commit to this branch.. it's tagged as `v3.3.1` with a corresponding crx under [releases](https://github.com/warren-bank/fork-crx-PrivacyPass-challenge-bypass/releases)

* makes a tiny change to the webpack configs
* the resulting crx:
  - is nearly half the size
  - contains ES5 code that will run in many more (ie: older) browsers
    * off=hand, I'd guess that support for `window.crypto.getRandomValues` is the bottleneck, but according to [this](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues).. it was available in Chrome v11

If there's a reason that you've chosen to target ES2020.. feel free to exclude this commit.

Annecodally, I tested the resulting crx and it works exactly the same.

README.md
warren-bank: another option.. which might be preferable.. would be to:
* move the `.bin` directory to `dist/.bin`
* configure webpack to output to `dist/PrivacyPass`
* configure `.gitignore` to exclude: `dist/PrivacyPass` and `dist/PrivacyPass.[pem|crx|xpi]`

so all output.. and my scripts to pack the output into extensions.. would be confined under `dist`

warren-bank:
2x point to bring up (before I forget)..
1. In the v2 extension, I'm pretty sure that the _hCaptcha_ provider defaulted to issuing 5x tokens per solved captcha because no explicit value was defined.. and 5x was inherited from the default configs. When configuring the new _hCaptcha_ provider, I set the value to 5x tokens.. because I didn't know if it would be OK to increase the number.. for example to 30x, as Cloudflare does. Do you know what limit is set on the server-side? ..whether it is 5.. or if we could go higher?
2. It would appear that the way _hCaptcha_ issues tokens prevents the ability to do so from within an incognito window; ..or rather, makes it more difficult.. and will require a workaround. My observation is that the user's POST containing the captcha solution (Req1) needs to occur before our POST to containing the tokens be signed (Req2). Furthermore, when Req1 occurs in an incognito window, Req2 always fails. My assumption is that the response to Req1 includes a cookie that needs to be sent back in Req2. That would explain why an incognito window (which sandboxes its cookies) would prevent Req2 (which is made by the background page and doesn't live within the incognito sandbox) from working. Is this something that you've already looked into.. and brainstormed workarounds for?

README.md
warren-bank: __update__:
That last suggestion seemed (to me) as a much cleaner option.. so I just pushed a commit to make those changes.

src/background/providers/cloudflare.ts
warren-bank: __update:__
Regarding my earlier comment/assertion that hCaptcha requires the timeout.. and also..   
Regarding my earlier comments about testing the ES5-transpiled release of the extension in older browsers..  
I just ran a very interesting test that addresses both topics:
* added a [polyfill library](https://cdnjs.cloudflare.com/ajax/libs/core-js/3.20.3/minified.min.js) to the _public_ directory and loaded it from both the background page and the popup
* using Chrome 30:
  - packaged the extension to a crx2 and installed
    * no errors or warnings.. everything looks good
  - solved an _hCaptcha_ and inspected network traffic
    * request from normal window tab to solve captcha: 200
    * request from background page to issue tokens: 403
      - `{"success":false,"error-codes":["invalid-data"]}`
      - which is the response that occurs when the request to issue tokens happens _before_ the POST to solve
* made a quick change to the timeout in the background page: increased from `0` to `2500` milliseconds
  - repacked the crx2 and reinstalled
  - solved another _hCaptcha_ and... earned 5x tokens
    * it works in Chrome 30! ..how cool is that?
* the Cloudflare provider doesn't work out-of-the-box
  - haven't looked into it yet
  - guessing it's because the TLS certificate isn't trusted by this browser
    * which would make the necessary `window.crypto` API unavailable
    * I'll need to add its root authority.. and retest..

warren-bank: I just added another commit to this branch.. it's tagged as `v3.4.0` with a corresponding crx under [releases](https://github.com/warren-bank/fork-crx-PrivacyPass-challenge-bypass/releases)

* adds hook functions that delegate to providers for:
  - [`chrome.webRequest.onCompleted`](https://developer.chrome.com/docs/extensions/reference/webRequest/#event-onCompleted)
  - [`chrome.webRequest.onErrorOccurred`](https://developer.chrome.com/docs/extensions/reference/webRequest/#event-onErrorOccurred)
* reverts the Cloudflare provider to remove the previously added timeout
* refactors the hCaptcha provider to move the code that sends issuing requests
  - from a timeout in 'handleBeforeRequest'
    * ie: the hook for [`chrome.webRequest.onBeforeRequest`](https://developer.chrome.com/docs/extensions/reference/webRequest/#event-onBeforeRequest)
  - to 'handleOnCompleted' and 'handleOnErrorOccurred':
    * ie: the new hooks for [`chrome.webRequest.onCompleted`](https://developer.chrome.com/docs/extensions/reference/webRequest/#event-onCompleted) and [`chrome.webRequest.onErrorOccurred`](https://developer.chrome.com/docs/extensions/reference/webRequest/#event-onErrorOccurred), respectively
* updates tests to adjust for revisions
  - all pass

this design is:
1. much cleaner
2. much more reliable, since the timing of the secondary request depends upon completion of the 1st

the result is:
* stable and fully functional

warren-bank: PS: this new release was built using the improved build scripts, which automates the insertion of the ES6+ polyfill library (that I mentioned previously) into the CRX2 extension.. so
* the CRX3 does _not_ include polyfill
  - since it only runs in newer browsers
* the CRX2 does include the polyfill
  - since it only runs in legacy browsers
  - this backports support for some crazy old browsers..
    * the primary bottleneck now is whether or not the website operated by each provider will also support the older browsers

MikeGitAZ:
On Windows 10 with Chrome 97.0.4692.99 this pull request increments the hCaptcha counter when I refill but the tokens don't do anything when I encounter an actual hCaptcha (i.e. I'm forced to solve the hCaptcha).

warren-bank:
@MikeGitAZ Holy cow! You're absolutely right. I wasn't aware of URLs for websites to test this on.. and just assumed that the same redemption logic would work for HC as CF. I just scoured the web looking for sites to test on.. and found:

Cloudflare:
* [streameast.live](https://streameast.live/)
  - works: access to the site is gated
    * no extension: must solve captcha to access lists of streams
    * with extension: bypass

hCaptcha:
* [aaep.org](https://aaep.org/contact)
  - does not work: the "contact us" form displays an "I am human" checkbox just above the "send message" button

That being said.. in the network tab, I can see the request that the extension _should_ be responding to.. and I can see exactly how I messed it up. I'll fix it shortly.

Thanks so much for the feedback! :)

MikeGitAZ:
@warren-bank No problem. Being outside of the U.S. temporarily, I'm forced to deal with these things way too often, so I'm really looking forward to when this extension is fixed!

warren-bank:
Hi @MikeGitAZ ..please give [v3.6.0](https://github.com/warren-bank/fork-crx-PrivacyPass-challenge-bypass/releases/tag/v3.6.0) a try. Redemption looks good on my end (with what limited testing I just did).

warren-bank:
to do:

* investigate why the changes made to the _hCaptcha_ provider (to make it work)..
  - specifically the relative timing of the 2x requests
  - wait for 1st to complete _before_ triggering the 2nd to issue new tokens
* are seemingly creating a race condition in the _Cloudflare_ provider for issuing new tokens
  - half the time it works
  - half the time it doesn't recognize the request it should
    * the request to intercept and cancel
    * that should immediately trigger a 2nd request to issue new tokens

warren-bank:
regarding restoration of JSON token backup data files.. to do:

* investigate a better method to open a file input dialog
  - works well in newer versions of Chrome
    * where the extension icons in the toolbar are grouped together under a single dropdown
  - in older versions of Chrome, I see mixed results..
    * the problem traces to the fact that the file input dialog causes the popup window to lose focus and close
    * this causes the script that is executing in the popup window (which opened the file input dialog) to abruptly terminate
    * the file that the user chooses in the dialog aren't processed.. because there is no callback function to do the work
    * a silly workaround for this is to open devtools for the popup window while restoring a token backup

MikeGitAZ:
Still not working. Here's what I see in the console:
`08:48:53.972 api.js?render=explicit&recaptchacompat=off&onload=_cf_chl_hload:2 recaptchacompat disabled
08:49:01.927 DevTools failed to load source map: Could not load content for chrome-extension://abkfbakhjpmblaafnpgjppbmioombali/lib/browser-polyfill.js.map: HTTP error: status code 404, net::ERR_UNKNOWN_URL_SCHEME
`

warren-bank:
fwiw, those 2x console log messages are not relevent..

* `api.js` is just a javascript file that the page loads
  - not involved in the 2-way interaction
* [`abkfbakhjpmblaafnpgjppbmioombali`](https://www.google.com/search?q=%22abkfbakhjpmblaafnpgjppbmioombali%22) corresponds to a different [open-source](https://github.com/WorldBrain/Memex) Chrome [extension](https://chrome.google.com/webstore/detail/memex/abkfbakhjpmblaafnpgjppbmioombali)

warren-bank:
Hi @MikeGitAZ ..please give [v3.6.1](https://github.com/warren-bank/fork-crx-PrivacyPass-challenge-bypass/releases/tag/v3.6.1) a try. The git commit message goes into detail.. but tl;dr.. @ppopth isn't going to like this.. but wrapping async code in a timer to delay execution until next tick is back.. and it makes Cloudflare token issuance much more reliable.. not perfect.. but almost. There's still a race-condition.. but this is a big improvement.

MikeGitAZ:
I should have reported this before, but the hCaptcha token count on the icon shows zero on sites other than the refill site (although the total is correct when I click the icon).

Anyway, it still isn't working and now I see this in the console:

> Uncaught (in promise) Error: No registered remote function called removeHighlights
    at h.<anonymous> (chrome-extension://abkfbakhjpmblaafnpgjppbmioombali/content_script.js:2:28401)
    at Generator.next (<anonymous>)
    at o (chrome-extension://abkfbakhjpmblaafnpgjppbmioombali/content_script.js:2:2746)

warren-bank:
@ppopth ..this race condition is owned by the _Cloudflare_ code on its website frontend.. which I don't have any insight into what logic it uses to submit the captcha data to its backend.

I see 2x types of requests being made.. do you have any idea what happens in the _Cloudflare_ code to trigger one vs. the other? The relative frequency doesn't appear to be consistent.. so I can't really say X percent of the time a _good_ request occurs. The shape of these 2x types of requests look like:

__bad request:__
```bash
curl 'https://captcha.website/' \
  -H 'content-type: application/x-www-form-urlencoded' \
  -H 'referer: https://captcha.website/?__cf_chl_tk=xxx' \
  --data-raw 'md=xxx&r=xxx&cf_captcha_kind=h&vc=xxx&captcha_vc=xxx&captcha_answer=xxx&cf_ch_verify=plat&h-captcha-response=captchka' \
```

__good request:__
```bash
curl 'https://captcha.website/?__cf_chl_captcha_tk__=xxx' \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -H 'Referer: https://captcha.website/' \
  --data-raw 'md=xxx&r=xxx&cf_captcha_kind=h&vc=xxx&captcha_vc=xxx&captcha_answer=xxx&cf_ch_verify=plat&h-captcha-response=captchka' \
```

warren-bank:
as a follow-up question..

the extension could easily detect the _bad_ request and process it as though it were good; however, the url in a _bad_ request doesn't include the `__cf_chl_captcha_tk__` querystring parameter..
1. does this querystring parameter need to be included in the url for the secondary request (ie: to issue tokens)?
   - if not.. then problem solved
   - if so.. then:
2. is the value of this querystring value the same as the value of the `__cf_chl_tk` querystring parameter in the 'referer' request header of a _bad_ request?
   - if so.. then problem solved (with some additional plumbing to access this value)
   - if not.. then can it be determined from a _bad_ request in some other way?

warren-bank:
I can actually already answer the first question.. because I've tested this approach.. and can say with certainty that the `__cf_chl_captcha_tk__` querystring parameter is required in the secondary request to issue tokens; without it, the server responds with a 403 code.

warren-bank:
I can now also answer the 2nd question..

I'm going to refer to this strategy as the "rage against the machine" approach, because:
* it's super aggressive
* it's pretty much saying to the frontend code: "f-you, I don't care what you tell me"

Now, that being said.. it works! Like perfectly.. 100% of the time.

I'm going to eat dinner first.. and then clean up the code a little and add a few new test cases before pushing this commit/release. Will be added shortly..

warren-bank:
[v3.6.2](https://github.com/warren-bank/fork-crx-PrivacyPass-challenge-bypass/releases/tag/v3.6.2) is available for test. The methodology may be debatable, but you can't argue with results; this version works 100% of the time (..for me, during my limited testing).

MikeGitAZ:
I guess the message in my previous post also referred to a different extension...my mistake. Anyway, here's the URL I'm testing hCaptcha functionality on without success: https://www.ancestry.com/search/collections/30105/

warren-bank:
@MikeGitAZ ..I can't reproduce. When I load your [URL](https://www.ancestry.com/search/collections/30105/), the page doesn't include code for any captcha providers. Maybe the backend conditionally includes scripts based on IP geolocation.. dunno? I'll test it later using a VPN tunnel from an international server.

warren-bank:
regarding issuing hCaptcha tokens in Firefox.. all I can say is: _WTF_?

* the request to issue tokens receives a 200 response about 5% of the time.. and tokens are issued
  - `{"status":200,"signatures":"xxx"}`
* the other 95% of the time.. the request receives a 403 response.. and no tokens are issued
  - `{"success":false,"error-codes":["invalid-data"]}`
* I compared 2x requests.. 1x that received a 200 and 1x that received a 403
  - there are no added or missing headers
  - aside from the value of the data payload, everything is identical
* the extension explicitly waits until the first request completes for issuing the request to issue tokens
  - I can see in the network tab that this first request always completes successfully
* in Chrome, this timing dependency is what allows the second request to always receive a 200 response
  - the same timing occurs in Firefox.. so I have no idea why the server would respond differently

fwiw, I was curious how `v2.0.9` behaved in Firefox.. and I see the exact same behavior; all requests to hCaptcha to issue tokens return a 403 status code. So.. (at least in Firefox 97-dev).. the older extension is broken as well.. for the same reason. Although.. that older extension didn't make any attempt to delay the 2nd request until after the 1st is completed.

MikeGitAZ:
@warren-bank Yep, that seems to be the case. When I use a US-based VPN server I don't get the captcha. I'm in Brazil; free VPN access through servers based here is difficult to come by.

warren-bank:
I have several non-free VPN accounts.. I'm sure finding a server to test on won't be a problem. I'll test from Brazil when I do.

MikeGitAZ:
> * the request to issue tokens receives a 200 response about 5% of the time.. and tokens are issued
>   
>   * `{"status":200,"signatures":"xxx"}`
> * the other 95% of the time.. the request receives a 403 response.. and no tokens are issued

This seems consistent with my personal hCaptcha success rate so the extension appears to be working... :D

warren-bank:
Are you using Firefox?

MikeGitAZ:
No, Chrome. That was half joking although I definitely get frustrated with the captchas at times.

warren-bank:
Oh ok. Well.. personally, I couldn't care less about Firefox.. but this is just super frustating.. because the extension appears to be doing everything right. Without knowing what's going on at the server.. it's just banging my head against a wall.

MikeGitAZ:
I understand. Personally, I was somewhat surprised to discover that such an extension/add-on could even have a chance to work. I figured that the captcha people wouldn't like the idea of it and would code defensively against such functionality, even though to me it doesn't seem like much of a threat. It's just that they're all about gatekeeping.

warren-bank:
From the first sentence of the [README](https://github.com/privacypass/challenge-bypass-extension#privacy-pass-extension):
> The Privacy Pass protocol is now being standardised by the [privacypass](https://datatracker.ietf.org/wg/privacypass/about/) IETF working group.

This project isn't about finding a clever way to work around the capcha providers. My understanding is that they want this solution to work, and are in collaboration to establish a standard. To do so doesn't make their services obsolete.. quite the opposite; it makes their services more transparent.. less invasive.. and more likely to be used ubiquitously across the interweb.

MikeGitAZ:
Yes, I get that, but I was rather surprised when I first found out about it. I'm just saying that security types tend to be on the cautious side.

warren-bank:
@MikeGitAZ ..I just tested the 3x URLs that have been listed in this conversation from a VPN server in São Paulo, Brazil:

* [https://www.ancestry.com/search/collections/30105/](https://www.ancestry.com/search/collections/30105/)
  - _success_
  - cloudflare: redeems 1x token per session (not per page load)
* [https://streameast.live/](https://streameast.live/)
  - _success_
  - cloudflare: redeems 1x token per session (not per page load)
* [https://aaep.org/contact](https://aaep.org/contact)
  - _failure_
  - this is interesting..
    * in the US: hCaptcha is the only provider on the page (checkbox in iframe)
      - hCaptcha plugin is active in extension
      - captcha is solved automatically by hCaptcha plugin when checkbox is clicked
    * in Brazil: in addition to hCaptcha, Cloudflare is detected on the page (though no tokens are redeemed during page load)
      - Cloudflare plugin is active in extension, because it is intentionally prioritized
      - captcha is __not__ solved automatically by hCaptcha plugin when checkbox is clicked, because the hCaptcha plugin is not active

warren-bank:
regarding issuing hCaptcha tokens in Firefox..
* I figured it out what the problem is!
  - it has nothing to do with the extension, which is working perfectly
  - it is 100% a Firefox security policy/configuration issue
* by default, Firefox blocks all 3rd-party cookies
  - apparently, this security policy carries over to xhr requests that are made by extensions in the background page
  - since hCaptcha requires 2x requests:
    * the 1st solves a captcha and (presumably) sets a cookie
    * the 2nd requests that tokens be signed, which (presumably) requires the presence of the cookie that is (presumably) set by the response to the 1st request
  - this methodology is very fragile and subject to various security policies:
    * in all browsers:
      - when the 1st request is made from a _private_ window, the cookie (presumably) set by its response is sandboxed.. and is never available to the 2nd xhr request made by the extension in its background page
    * in Firefox:
      - when 3rd-party cookies are blocked.. then regardless of whether the 1st request is made from a _private_ or _normal_ window, the cookie (presumably) set by its response is never sent by the 2nd xhr request made by the extension in its background page.. because the "origin" of the background page isn't 1st-party
      - the attached screenshot shows the configuration change that I just made that results in the extension fully working 100% of the time in Firefox to issue tokens from a _normal_ window..

Firefox-screenshot

warren-bank:
disregard my last comment..
* I thought I found the underlying cause of the problem with hCaptcha in Firefox (re: blocked 3rd-party cookie)
* it seemed to be working perfectly during testing
  - which annecdotally always seems to be the case immediately after I restart Firefox
* old/broken behavior a few minutes later

warren-bank:
summary of unresolved issues:
* all browsers:
  - hCaptcha
    * cannot issue tokens from a private/incognito window
  - Cloudflare
    * cannot issue tokens when a "User-Agent" switcher extension is active
* Firefox:
  - hCaptcha
    * ability to issue tokens (from a normal window) is erratic

ppopth:
Hi @warren-bank, Since this PR is quite huge and it touches the code structure of the project which needs discussion, would you mind if I ask you to create a smaller PR first? (For example, the PR which only makes hCaptcha tokens work)

warren-bank:
> Since this PR is quite huge and it touches the code structure of the project which needs discussion, would you mind if I ask you to create a smaller PR first? (For example, the PR which only makes hCaptcha tokens work)

Though I have noticed that a few of my more minor tweaks have already been cherry picked into standalone PRs that have already been merged.. the trouble with this request is that none of the more serious aspects of the code can live in a sandbox, isolated from the rest of the project.

Using the _hCaptcha_ provider plugin as an example:
* it requires hooks to listeners for `chrome.webRequest` events that don't exist
* it requires a library of common tools that doesn't exist (and much earlier, you voiced dislike for the idea of having)
* ...and probably many other things that don't immediately come to mind, as I haven't looked at this code in quite a while

> Another one is could you make the commit message up-to-date? for example, "work in progess.." means that the work is not finished yet.
>
> There are many things that should be done by the maintainer, such as updating the version number.

These commit messages and version numbers were made during a period of very active development in a separate fork. I never assumed that the dev branch would be merged verbatim. If you were to perform a squashed merge, then:
* all of the commit messages (both the way _way_ overly descriptive ones, as well as non-descriptive ones) will go away
* the version number can be rolled back to fit the official numbering chronology prior to commit

My point is.. that none of this should matter.

armfazh:
> Do you think it would be easier to discuss and selectively undo the parts you don't like?

What if we add the parts incrementally with small PRs.
So far, there are a lot of changes and they are difficult to track and review (not saying they are not useful). 

I can spot some features that are, by themselves, good additions for the extension:
 - Localization engine.
 - Browser detection
 - cookies handler
 - Any other UI change.
 - hCaptcha support (this may be a big change too).

@warren-bank would you like to create PRs addressing specific changes, or are you ok if we cherry pick changes from your fork? 
We appreciate your input and willingness to collaborate :)

warren-bank:
> would you like to create PRs addressing specific changes, or are you ok if we cherry pick changes from your fork?

you can cherry pick code from the fork, if:
* you add back the same LICENSE to your repo that was present @ [v2.0.9](https://github.com/privacypass/challenge-bypass-extension/blob/v2.0.9/LICENSE), which has since been removed
* you're willing to add me to your list of [contributors](https://github.com/warren-bank/fork-crx-PrivacyPass-challenge-bypass/blob/v3.6.6/package.json#L8)

regarding your ability to incrementally add features, rather than lump sum..<br>I'd highly recommend that you inspect the code diff for each commit in this PR in chronological order..<br>in addition to the contemporaneous comments/thoughts/questions written to the "conversation" section of this PR;
most features were added one at a time, and build on previous changes/additions/enhancements.
you might be able to say something like:
* commits a-b are PR 283.1
* commits b-c are PR 283.2
* etc..

> We appreciate your input and willingness to collaborate

I'm absolutely willing/eager to collaborate;<br>I'm just leary of being asked to start over and rewrite history..<br>I'd be much more interested in collectively trying to figure out how to fix/workaround the [remaining issues..](#issuecomment-1025258736)
so the extension's behavior is consistent and reliable for all users

Thaddeuswider55555th (approved):

PR #285, opened 2022-01-18
Resolve path correctly.

### LINKED ISSUE #282: webpack config doesn't build on Windows
the declaration of the variable [`__dirname`](https://github.com/privacypass/challenge-bypass-extension/blob/266a560e9650118e739586dcb339b8cdfeee790d/webpack.config.js#L10) currently assumes `path.sep` is `/`.

a more platform agnostic declaration would look more like:

```javascript
const __dirname = (() => {
  const filepath_uri = import.meta.url;
  const prefix = `file:${path.sep === '/' ? '' : path.sep}`;
  let fp;
  fp = path.normalize(filepath_uri);
  fp = (fp.indexOf(prefix) === 0) ? fp.substring(prefix.length, fp.length) : new URL(filepath_uri).pathname;
  fp = path.dirname(fp);
  return fp;
})();
```

armfazh:
Fixes #282

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -2,12 +2,13 @@ import CopyWebpackPlugin from 'copy-webpack-plugin';
 import HtmlWebpackPlugin from 'html-webpack-plugin';
 import MiniCssExtractPlugin from 'mini-css-extract-plugin';
 import TsconfigPathsPlugin from 'tsconfig-paths-webpack-plugin';
+import { fileURLToPath } from 'url';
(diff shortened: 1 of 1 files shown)


armfazh:
@warren-bank please confirm if this works ok in Windows.

warren-bank:
it does; I've been using that line of code all morning.. thanks for fixing :)

ppopth (approved):

Issue #291, opened 2022-02-02
Cloudflare Passes completely broken


RealDolos:
The reason for this seems simple: you (cloudflare) rolled out a new version of the captcha challenge that broke your (privacy pass) extension.

E.g. the extension looks for POSTs with `__cf_chl_captcha_tk__'  or '__cf_chl_managed_tk__` to detect a challenge submission ([see](https://github.com/privacypass/challenge-bypass-extension/blob/6c0c3a80eb46878a6629de12c61aac4fc180ea5c/src/background/providers/cloudflare.ts#L240)). But requests with these parameters do not exist anymore. The closest is `__cf_chl_rt_tk`, but that one isn't even used in the POST itself, it's just pushed onto the browser history when that challenge js is loaded and replaced again afterwards. The final submission POST cannot be detected that way, and the extension will therefore never even attempt to do an issuance.

Tested broken: Firefox and Chrome (no surprise), extension version 3.0.0

ppopth:
#292 for tracking

mendelsphotography:
By me had the same issue but seemed to be by me it screwed up the CSS but if used in Incognito it was fine and as soon I disabled the extension (privacy pass) soo don't know if that's related also or just me

ppopth:
This has been resolved in v3.0.1

PR #292, opened 2022-02-02
Change cloudflare issue request flow

ppopth:
Since we don't have any query param in the issue request any more, we remove the code which uses the query params to identify the issue request.

Currently we have a Referer header. We also need to use that header in the issue request as well. However such change requires a change in the code structure, as follows:

Previously we used axios to send an issuance request in handleBeforeRequest, but now we will send the request in
handleBeforeSendHeaders because we need to read the Referer header first.

After reading the Referer header, we extract the __cf_chl_tk query param in the Referer url and add the __cf_chl_f_tk query param with that token in the issuance request URL.

We need a new "issueInfo" property to remember the request body we can read in handleBeforeRequest and then it will be read in handleBeforeSendHeaders so that in handleBeforeSendHeaders we will have both the body and the Referer header.

### DIFF
--- a/src/background/index.ts
+++ b/src/background/index.ts
@@ -21,8 +21,27 @@ declare global {
     }
 }
 
+declare let browser: any;
+
 window.ACTIVE_TAB_ID = chrome.tabs.TAB_ID_NONE;
 window.TABS = new Map<number, Tab>();
+
+const BROWSERS = {
+    CHROME: 'Chrome',
+    FIREFOX: 'Firefox',
+    EDGE: 'Edge',
+} as const;
+type BROWSERS = typeof BROWSERS[keyof typeof BROWSERS];
+
+function getBrowser(): BROWSERS {
+    if (typeof chrome !== 'undefined') {
+        if (typeof browser !== 'undefined') {
+            return BROWSERS.FIREFOX;
+        }
+        return BROWSERS.CHROME;
+    }
+    return BROWSERS.EDGE;
+}
 
 /* Listeners for navigator */
 
@@ -57,10 +76,14 @@ chrome.webRequest.onBeforeRequest.addListener(handleBeforeRequest, { urls: ['<al
     'blocking',
 ]);
 
+const extraInfos = ['requestHeaders', 'blocking'];
+if (getBrowser() === BROWSERS.CHROME) {
+    extraInfos.push('extraHeaders');
+}
 chrome.webRequest.onBeforeSendHeaders.addListener(
     handleBeforeSendHeaders,
     { urls: ['<all_urls>'] },
-    ['requestHeaders', 'blocking'],
+    extraInfos,
 );
 
 chrome.webRequest.onHeadersReceived.addListener(handleHeadersReceived, { urls: ['<all_urls>'] }, [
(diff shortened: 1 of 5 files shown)


src/background/index.ts
```diff
@@ -21,9 +21,23 @@ declare global {
...
+declare let browser: any;
+
 window.ACTIVE_TAB_ID = chrome.tabs.TAB_ID_NONE;
 window.TABS = new Map<number, Tab>();
 
+function getBrowser() {
+    if (typeof chrome !== 'undefined') {
+        if (typeof browser !== 'undefined') {
+            return 'Firefox';
+        } else {
```
thibmeu, line 33: no else after return

src/background/index.ts
```diff
@@ -21,9 +21,23 @@ declare global {
...
+            return 'Chrome';
+        }
+    } else {
```
thibmeu, line 36: no else after return

src/background/index.ts
```diff
@@ -57,10 +71,14 @@ chrome.webRequest.onBeforeRequest.addListener(handleBeforeRequest, { urls: ['<al
     'blocking',
 ]);
 
+const extraInfos = ['requestHeaders', 'blocking'];
+if (getBrowser() === 'Chrome') {
```
thibmeu, line 75: maybe we can have a const dicvtionary with browser
something like
```typescript
const BROWSERS = {
  CHROME: 'Chrome',
  FIREFOX: 'Firefox',
  EDGE: 'Edge',
} as const;
type BROWSERS = typeof BROWSERS[keyof typeof BROWSERS];
```

ppopth: I changed to enum instead then.

thibmeu: `enum` is not well built in typescript, and can rapidly makes code hard to read. Using a const map works better.

src/background/providers/cloudflare.test.ts
```diff
@@ -116,17 +215,58 @@ describe('issuance', () => {
...
+                timeStamp: 1,
+                requestHeaders: [],
+            };
+            const result = await provider.handleBeforeSendHeaders(details);
+            expect(result).toStrictEqual({ cancel: true });
+            const newIssueInfo = provider['issueInfo'];
+            expect(newIssueInfo).toBeNull();
+
+            expect(issue.mock.calls.length).toBe(1);
+            expect(issue).toHaveBeenCalledWith('https://captcha.website/?__cf_chl_f_tk=token', {
```
migueldemoura, line 254: Should we check that this request's `Referer` header will either not be present or contain the URL without the parameter?

migueldemoura: I meant for the challenge solve request, and the `Referer` not being present is perfect. Thanks!

RealDolos:
fwiw, with this PR things seem to work in Firefox Nightly. Thanks

armfazh (approved):

mendelsphotography:
Does this fix the issue #291 or no? Just wondering since just happened to me the other day and I tught it was me since I reinstalled chrome.
Because the lastest release is  from a long time ago so does that mean this need to be compiled on my PC now?

warren-bank:
fwiw, you might want to compare the differences in implementation between this PR and PR #283 in the 2x commits made on 01/29/2022. The code here uses some very loose browser detection, whereas the code in the other PR uses feature detection.. which will support older versions of all supported browsers. For example, the code here will break in versions of Chrome < 72.

PR #303, opened 2022-02-15
Fixes some issues reported by linter.

armfazh:


src/background/providers/cloudflare.test.ts
```diff
@@ -40,8 +40,11 @@ test('setStoredTokens', () => {
     const provider = new CloudflareProvider(storage, { updateIcon, navigateUrl });
     const tokens = [new Token(), new Token()];
     provider['setStoredTokens'](tokens);
-    const storedTokens = JSON.parse(storage.store.get('tokens')!);
-    expect(storedTokens).toEqual(tokens.map((token) => token.toString()));
+    const tok = storage.store.get('tokens');
+    if (tok) {
```
ppopth, line 44: Could this be more explicit whether it's checking for `false`, `null`, or `undefined`?

src/background/providers/cloudflare.test.ts
```diff
@@ -117,11 +120,13 @@ describe('issuance', () => {
             expect(navigateUrl).not.toHaveBeenCalled();
 
             const issueInfo = provider['issueInfo'];
-            expect(issueInfo!.requestId).toEqual(details.requestId);
-            expect(issueInfo!.formData).toStrictEqual({
-                ['h-captcha-response']: 'body-param',
-                ['cf_captcha_kind']: 'body-param',
-            });
+            if (issueInfo) {
```
ppopth, line 123: Could this be more explicit whether it's checking for `false`, `null`, or `undefined`?

src/background/providers/cloudflare.test.ts
```diff
@@ -355,8 +360,10 @@ describe('redemption', () => {
             expect(result).toEqual({ redirectUrl: details.url });
             // Expect redeemInfo to be set.
             const redeemInfo = provider['redeemInfo'];
-            expect(redeemInfo!.requestId).toEqual(details.requestId);
-            expect(redeemInfo!.token.toString()).toEqual(tokens[0].toString());
+            if (redeemInfo) {
```
ppopth, line 363: Could this be more explicit whether it's checking for `false`, `null`, or `undefined`?

src/background/providers/cloudflare.ts
```diff
@@ -242,21 +243,21 @@ export class CloudflareProvider implements Provider {
             // header according to https://xhr.spec.whatwg.org/#dom-xmlhttprequest-setrequestheader
             // So we need to extract the token from the Referer header and send it in the query
             // param __cf_chl_f_tk instead. (Note that this token is not a Privacy Pass token.
-            let token: string | null = null;
+            let atoken: string | null = null;
```
ppopth, line 246: What does this mean? Is it a lint error? I cannot find it from `npm run lint`?

src/background/providers/cloudflare.ts
```diff
@@ -335,14 +340,14 @@ export class CloudflareProvider implements Provider {
 
         // Get one token.
         const tokens = this.getStoredTokens();
-        const token = tokens.shift();
+        const oneToken = tokens.shift();
```
ppopth, line 343: I think `token` (without s) already means one token.

src/background/providers/cloudflare.ts
ppopth: I think `token` (without s) already means a token.

src/background/providers/cloudflare.ts
```diff
@@ -288,19 +289,23 @@ export class CloudflareProvider implements Provider {
...
         if (!hasBodyParams) {
             return;
         }
 
         const flattenFormData: { [key: string]: string[] | string } = {};
         for (const key in details.requestBody.formData) {
-            if (details.requestBody.formData[key].length == 1) {
-                const [value] = details.requestBody.formData[key];
-                flattenFormData[key] = value;
+            if (details.requestBody.formData[key as string].length == 1) {
```
ppopth, line 304: This seems to be a vulnerability rather than a lint error. https://github.com/nodesecurity/eslint-plugin-security/blob/master/docs/the-dangers-of-square-bracket-notation.md

This patch make the linter pass rather fixing the vulnerability. https://stackoverflow.com/questions/51273469/how-to-resolve-eslint-generic-object-injection-sink-error

ppopth: I think defining a function checking for the key of `constructor`, `__proto__`, and `prototype` should be fine.

src/background/providers/cloudflare.ts
armfazh: the issue is that the linter looks for variables named `token` or `password` and warns about it. The fix proposed is to rename the variable.

src/background/providers/cloudflare.ts
armfazh: Linter reports `security/detect-possible-timing-attacks` when detects a variable named `token` or `password`. 

https://github.com/gkouziik/eslint-plugin-security-node/blob/master/lib/rules/detect-possible-timing-attacks.js

src/background/providers/cloudflare.test.ts
armfazh: done

src/background/providers/cloudflare.test.ts
armfazh: done

src/background/providers/cloudflare.ts
```diff
@@ -288,19 +289,23 @@ export class CloudflareProvider implements Provider {
         }
 
         const hasBodyParams = QUALIFIED_BODY_PARAMS.every((param) => {
-            return details.requestBody !== null && param in details.requestBody.formData!;
+            return (
```
ppopth, line 292: I think it's better to be more specific what you are checking for. Like check for `null`. This line gives a hint to me that `requestBody` may be boolean.

ppopth (approved):

src/background/providers/cloudflare.ts
armfazh: agree, fixed now

PR #344, opened 2022-08-31
restore hCaptcha compatibility

fedecarpy:
This PR restores full hCaptcha compatibility.

Maintainers: please notify support@hcaptcha.com well in advance of shipping breaking changes to hCaptcha support. This will let us schedule work to validate and PR any changes required.

We have historically linked hCaptcha users to the extension published out of this repo, but if maintaining compatibility with non-Cloudflare endpoints is not a priority then it may make more sense to publish our fork separately to avoid similar issues going forward.

Let us know your thoughts on this.


src/background/providers/hcaptcha.ts
```diff
@@ -24,19 +81,179 @@ export class HcaptchaProvider implements Provider {
         this.callbacks.updateIcon(this.getBadgeText());
     }
 
+    private handleUrl(url: URL) {
+        const reqUrl = url.origin + url.pathname;
+        const isIssuerUrl = !!reqUrl.match(ISSUER_REGEX);
```
thibmeu, line 86:
```suggestion
        const isIssuerUrl = ISSUER_REGEX.test(reqUrl);
```

src/background/providers/hcaptcha.ts
```diff
@@ -24,19 +81,179 @@ export class HcaptchaProvider implements Provider {
...
+
+        let isSpendUrl;
+        if (HCAPTCHA_URLS.some((url: string) => reqUrl === url)) {
```
thibmeu, line 89: to ease reading
```suggestion
        if (HCAPTCHA_URLS.includes(reqUrl)) {
```

you could even use
```suggestion
// test if the URL is not a special hCaptcha url, and if it a valid spend URL.
const isSpendUrl = !HCAPTCHA_URLS.includes(reqUrl) && SPEND_REGEX.test(reqUrl)
```

src/background/providers/hcaptcha.ts
```diff
@@ -24,19 +81,179 @@ export class HcaptchaProvider implements Provider {
...
+        // Download the commitment
+        const { data } = await axios.get<Response>(COMMITMENT_URL);
+        const commitment = data.HC[version as string];
+        if (commitment === undefined) {
+            throw new Error(`No commitment for the version ${version} is found`);
+        }
+
+        // Cache.
+        const item = {
+            G: commitment.G || voprf.sec1EncodeToBase64(voprf.getActiveECSettings().curve.G, false),
```
thibmeu, line 122: given `G` is a `string`, I think you are looking for `??`
```suggestion
            G: commitment.G ?? voprf.sec1EncodeToBase64(voprf.getActiveECSettings().curve.G, false),
```

src/background/providers/hcaptcha.ts
```diff
@@ -24,19 +81,179 @@ export class HcaptchaProvider implements Provider {
...
+        return {
+            reqUrl,
+            isIssuerUrl,
+            isSpendUrl,
+        };
+    }
+
+    private async getCommitment(version: string): Promise<{ G: string; H: string }> {
+        const keyPrefix = 'commitment-';
+        const cached = this.storage.getItem(`${keyPrefix}${version}`);
```
thibmeu, line 104: consider declaring
```
const key = `${keyPrefix}${version}`
``` as you use it multiple time

File: src/background/providers/hcaptcha.ts [hunk 5]
```diff
@@ -24,19 +81,179 @@ export class HcaptchaProvider implements Provider {
...
+            // Do not store infinite tokens
+            const tokens = this.getStoredTokens();
+            if (tokens.length + NUM_OF_REQ_TOKENS > MAX_NUM_OF_TOKENS) return;
+
+            this.issueInfo = {
+                newUrl: details.url,
+                tabId: details.tabId,
+            };
+            return;
+        } else if (urlType.isSpendUrl) {
```
thibmeu, line 209: not else after return

src/background/providers/hcaptcha.ts
```diff
@@ -24,19 +81,179 @@ export class HcaptchaProvider implements Provider {
...
+            // Get one token
+            const tokens = this.getStoredTokens();
+            const oneToken = tokens.shift();
+            if (oneToken === undefined) return;
+            this.setStoredTokens(tokens);
+            this.forceUpdateIcon();
+
+            // Spend logic here
+            const httpPath = details.method + ' ' + url.pathname;
```
thibmeu, line 218: I feel this is a place where string interpolation is more readable.
```suggestion
            const httpPath = `${details.method} ${url.pathname}`;
```

src/background/providers/hcaptcha.ts
```diff
@@ -24,19 +81,179 @@ export class HcaptchaProvider implements Provider {
...
-        _details: chrome.webRequest.WebResponseHeadersDetails,
+        details: chrome.webRequest.WebResponseHeadersDetails,
     ): chrome.webRequest.BlockingResponse | void {
+        if (details.responseHeaders === undefined) return;
+        if (this.issueInfo === null) return;
+        if (details.tabId !== this.issueInfo.tabId) return;
+
+        const url = new URL(details.url);
+        const urlType = this.handleUrl(url);
+        if (!urlType.isIssuerUrl || details.statusCode === 403) return;
```
thibmeu, line 252: given preconditions are checked before calling `this.handleUrl()`, I think `details.statusCode === 403` should move at the same level.

src/background/providers/hcaptcha.ts
```diff
@@ -1,19 +1,76 @@
+import * as voprf from '../voprf';
```
thibmeu, line 1: there should be a test file associated to this change, similar to `cloudflare.test.ts`

src/background/providers/hcaptcha.ts
```diff
@@ -24,19 +81,179 @@ export class HcaptchaProvider implements Provider {
...
+        const item = {
+            G: commitment.G || voprf.sec1EncodeToBase64(voprf.getActiveECSettings().curve.G, false),
+            H: commitment.H,
+        };
+        this.storage.setItem(`${keyPrefix}${version}`, JSON.stringify(item));
+        return item;
+    }
+
+    private issue(url: string) {
+        (async () => {
```
thibmeu, line 130: what is the reason you declare an anonymous function to call it immediately, instead of making the private method `async`?

src/background/providers/hcaptcha.ts
```diff
@@ -24,19 +81,179 @@ export class HcaptchaProvider implements Provider {
...
+            const newTokens = Array.from(Array(NUM_OF_REQ_TOKENS).keys()).map(() => new Token());
```
thibmeu, line 131: improves readability
```suggestion
            const newTokens = Array(NUM_OF_REQ_TOKENS).fill(new Token());
```

src/background/voprf.js
```diff
@@ -55,14 +57,14 @@ export function initECSettings(h2cParams) {
     const methodStr = h2cParams.method;
     switch (curveStr) {
         case 'p256':
-            if (methodStr != 'swu' && methodStr != 'increment') {
+            if (methodStr !== 'swu' && methodStr !== 'increment') {
```
thibmeu, line 60: nice improvement

src/background/voprf.js
```diff
@@ -9,15 +9,17 @@ import 'asn1-parser';
 
 import sjcl from 'sjcl';
 
+import createKeccakHash from 'keccak';
```
thibmeu, line 12: consider using [`sha3`](https://www.npmjs.com/package/sha3), which does not rely on Node `Buffer` library which needs to be polyfilled.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "privacy-pass",
-    "version": "3.0.3",
+    "version": "3.0.4",
```
thibmeu, line 3: version bump should be in a different commit/PR

package.json
```diff
@@ -24,7 +24,7 @@
         "asn1-parser": "1.1.8",
         "axios": "0.23.0",
         "buffer": "6.0.3",
-        "keccak": "3.0.1",
+        "keccak": "3.0.2",
```
thibmeu, line 27: what's the reason to upgrade `keccak`?

thibmeu (commented):
the changes look good to me. the one thing I would like to see is a test file, even if it's basic, as it would avoid breaking hcaptcha compatibility in future upgrades

cjpatton:
Hi @fedecarpy, apologies for the delay on this. We intend to land this and will work with you on getting it merged in the next week or so.

First starters, we have some questions about maintenance, and it would be great to confirm that you actually work for hCaptcha :) Would you mind reach out at https://research.cloudflare.com/contact/?

package.json
fedecarpy: version 3.0.2 replaces `stream` with `readable-stream` so it doesn't require a polyfill

src/background/providers/hcaptcha.ts
fedecarpy: added a basic test file in this path `src/background/providers/hcaptcha.test.ts` 👍

src/background/providers/hcaptcha.ts
fedecarpy: removed!

src/background/providers/hcaptcha.ts
```diff
@@ -1,19 +1,76 @@
...
+import Token from '../token';
+import { Storage } from '../storage';
+import axios from 'axios';
+import qs from 'qs';
+
+const COMMITMENT_URL =
+    'https://raw.githubusercontent.com/privacypass/ec-commitments/master/commitments-p256.json';
+const SPEND_REGEX = RegExp('^https:\\/\\/(.+\\.)*hcaptcha.com\\/getcaptcha\\/(.*)$');
+const ISSUER_REGEX = RegExp('^https:\\/\\/(.+\\.)*hcaptcha.com\\/checkcaptcha\\/(.*)$');
+const HCAPTCHA_URLS = [
```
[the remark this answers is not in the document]
fedecarpy: yeah, that name is more clear for the future readers.
changed in last update 👍

src/background/providers/hcaptcha.ts
```diff
@@ -1,19 +1,76 @@
...
+    'https://hcaptcha.com/getcaptcha/00000000-0000-0000-0000-000000000000',
+    'https://hcaptcha.com/getcaptcha/10000000-ffff-ffff-ffff-000000000001',
+    'https://hcaptcha.com/getcaptcha/20000000-ffff-ffff-ffff-000000000002',
+    'https://hcaptcha.com/getcaptcha/30000000-ffff-ffff-ffff-000000000003',
+];
+
+const NUM_OF_REQ_TOKENS = 5;
```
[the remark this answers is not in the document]
fedecarpy: done 😃

src/background/providers/hcaptcha.ts
fedecarpy: nice suggestion, added to the PR

src/background/providers/hcaptcha.ts
fedecarpy: done!

src/background/providers/hcaptcha.ts
fedecarpy: `handleHeadersReceived` is a sync function and linter did not like leaving the issue function without await or .then()

changed the logic a little, tell me if you think is better that way 🤔

src/background/voprf.js
fedecarpy: the keccak package exposes the `squeeze` function that is used as a prng. sha3 does not, and even if i expose the squeeze function of that package and call it multiple times it just return the same hash 🤔 
could you suggest an implementation of that package please 😃

thibmeu:
@fedecarpy apologies for the delay. I'll add some more details on next week.

src/background/voprf.js
armfazh: let's continue using keccak.

armfazh (approved):
I have reviewed code and it is a neat implementation. 
I added one comment below.

armfazh:
Hi, I noticed that the hCaptcha allows to spend one token, but in the second try it consumes the token but do not bypass the captcha (the puzzle images are prompted to the user). See the image:


captchaOnceOnly

Steps to reproduce:
1. Get 5 tokens from https://www.hcaptcha.com/privacy-pass.
2. Go to https://accounts.hcaptcha.com/demo, in the first try, the token is consumed and no images. Good!
3. Refresh the page.
4. Check the mark of captcha, now, the token is consumed and sent without error, but the images still appear. :(

fedecarpy:
> Hi @fedecarpy , I noticed that hCaptcha allows to spend one token, but in the second try, it consumes the token but do not bypass the captcha (the puzzle images are prompted to the user). 

Hi @armfazh.

This is the expected behavior. The existence of a Privacy Pass token is only one of many variables that decide whether a challenge is shown. Rapid repeated redemptions make it less likely that a session will bypass the challenge.

src/background/providers/hcaptcha.ts
```diff
@@ -24,19 +81,172 @@ export class HcaptchaProvider implements Provider {
...
-        _details: chrome.webRequest.WebResponseHeadersDetails,
+        details: chrome.webRequest.WebResponseHeadersDetails,
     ): chrome.webRequest.BlockingResponse | void {
+        const url = new URL(details.url);
+        const urlType = this.handleUrl(url);
+        // wrong url or invalid status code
+        if (!urlType.isIssuerUrl || details.statusCode === 403) return;
+
+        // issueInfo was not loaded or the tab changed
+        if (this.issueInfo === null || details.tabId !== this.issueInfo.tabId) return;
```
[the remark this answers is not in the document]
fedecarpy: I do not think that it's necessary.
the _issueInfo_ is updated every time you call a valid IssuerUrl (and most likely the same one) and is cleared right after issue() is called a couple of lines later on that function.
And if you are spamming the endpoint, you will most likely not get any new tokens as there is double spend protection on our side.

armfazh:
> > Hi @fedecarpy , I noticed that hCaptcha allows to spend one token, but in the second try, it consumes the token but do not bypass the captcha (the puzzle images are prompted to the user).
> 
> This is the expected behavior. The existence of a Privacy Pass token is only one of many variables that decide whether a challenge is shown. Rapid repeated redemptions make it less likely that a session will bypass the challenge.

Ok, then we might add somewhere in the documentation that the captcha might still showing the puzzle even though the extension provides a valid token.
Tracked in #348

armfazh:
Thanks @fedecarpy , we will publish a new version soon.

PR #388, opened 2023-02-14
Remove param r in tests

armfazh:
**Changes**
UI: increase font for version number
type: remove any type
tests: remove parameter `r`

### DIFF
--- a/src/background/providers/cloudflare.test.ts
+++ b/src/background/providers/cloudflare.test.ts
@@ -111,7 +111,6 @@ describe('issuance', () => {
                 requestBody: {
                     formData: {
                         ['md']: ['body-param'],
-                        ['r']: ['body-param'],
                     },
                 },
             };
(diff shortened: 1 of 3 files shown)


src/background/providers/cloudflare.test.ts
```diff
@@ -154,7 +152,7 @@ describe('issuance', () => {
                 timeStamp: 1,
                 requestBody: {
                     formData: {
-                        ['md']: ['body-param'],
+                        /* remove 'md' parameter. */
```
migueldemoura, line 155: This test's comment says:
```
The request is invalid only if the body has both
         * 'md' and 'r' params.
```
Should we amend the comment? Also, what is the purpose of the test if we're not checking for both params any more?

armfazh: The amended comment for this test should be: 
`The request is invalid if the body has no 'md' parameter.`
will update it and the test. Thanks, well noted.

migueldemoura (approved):

PR #397, opened 2023-08-04
Fix Cloudflare tokens issuance domain

thibmeu:
Move from captcha.website to issuance.privacypass.cloudflare.com. This makes Cloudflare more identifiable as an issuer for Cloudflare tokens.

### DIFF
--- a/src/background/index.ts
+++ b/src/background/index.ts
@@ -113,11 +113,13 @@ chrome.runtime.onMessage.addListener((request, _sender, sendResponse) => {
 // TODO It's better to move this to the provider class. Let's figure out how to do it later.
 // Removes cookies for captcha.website to enable getting more tokens in the future.
 chrome.cookies.onChanged.addListener((changeInfo) => {
+    const cloudflareDomains = ['captcha.website', 'issuance.privacypass.cloudflare.com'];
+    const domain = changeInfo.cookie.domain.replace(/^\./, '');
     if (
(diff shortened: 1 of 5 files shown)


migueldemoura (approved):

Lekensteyn (commented):
The tests still list the old domain, should those also be updated?

src/background/index.ts
```diff
@@ -113,11 +113,14 @@ chrome.runtime.onMessage.addListener((request, _sender, sendResponse) => {
...
-        !changeInfo.removed &&
-        changeInfo.cookie.domain === '.captcha.website' &&
-        changeInfo.cookie.name === 'cf_clearance'
-    ) {
-        chrome.cookies.remove({ url: 'https://captcha.website', name: 'cf_clearance' });
+    const cloudflareDomains = ['captcha.website', 'issuance.privacypass.cloudflare.com']
+    for (const domain of cloudflareDomains) {
+        if (
+            !changeInfo.removed &&
+            changeInfo.cookie.domain === `.${domain}` &&
```
Lekensteyn, line 120: What about removing the outer loop:
```suggestion
            cloudflareDomains.includes(domain) &&
```
where `cloudflareDomains` is modified to have a period before?

Lekensteyn (approved):
Verified locally on Firefox 115.0.3 in a new browsing profile.

PR #398, opened 2023-08-30
Release v3.0.6

thibmeu:
- Change Cloudflare token retrieval domain
- Update dependencies
- Increase label font size


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "privacy-pass",
-    "version": "3.0.5",
+    "version": "3.0.6",
```
migueldemoura, line 3: Do you also need to bump the version here: https://github.com/privacypass/challenge-bypass-extension/blob/master/public/manifest.json#L5?

thibmeu: updated

migueldemoura (approved):

armfazh (approved):

PR #414, opened 2024-02-26
Add notice for end of support

thibmeu:
Privacy Pass has changed since the original PoPETS version. Moreover, CAPTCHA providers have moved from that version to follow IETF version more closely. They don't operate with a stored token model, but with an interactive token issuance and redemption, with a split attester/issuer model.
This commit updates the README to highlight the end of active token issuer/redeemer for this specific version of the protocol.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,9 @@
+# Deprecated
+
+#### This version of Privacy Pass extension is not currently under active development. More [details below](#deprecation).
+
+---
+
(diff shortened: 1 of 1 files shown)


migueldemoura (approved):

armfazh (approved):
This note in the timeline is ok, but also add a banner in the top of the README.

